PR #10, opened 2017-07-28
Feature/govendor experiment

sshaman1101:
Fix vendoring, now we must use govendor to manage external libs
successfully build on linux64 and mac os


Sokel:
`make install` instruction doesn't work

sshaman1101:
@Sokel now `make build && sudo make install` works correctly with typical Linux distro and macOS

jorpic:
**`sudo`**` make install`? rly?
Why not to use `$GOPATH/bin` as a default installation destination instead of `/usr/local/bin`?

sshaman1101:
My mistake, forgot about `$GOPATH/bin`, fixed

PR #12, opened 2017-07-28
WIP: Feature/add contracts

jorpic:
Please do not merge, this is still work in progress. But don't hesitate to comment and discuss.

This is an example of how generation of contract wrappers can be done via `go generate` tool.

This approach has some benefits:

  - no need to install `abigen` on developers machine as it is imported from `go-ethereum` by [contracts/generate-api.go](https://github.com/sonm-io/core/blob/ba6739b0ad930a722532efd96d841442648a4581/contracts/generate-api.go)
  - no need to store generated code in repo

Drawbacks:

  - requires `truffle` to be installed on developers machine

What is done:

  - contracts copied from `feature/invoke_blockchin_into_core` to `contracts` dir
  - npm/yarn is used to fetch solidity dependencies (OpenZeppelin)
  - contract wrappers are generated by `go generate` using code in [contracts/generate-api.go](https://github.com/sonm-io/core/blob/ba6739b0ad930a722532efd96d841442648a4581/contracts/generate-api.go)

What is still to be done:

  - migrate tests from blockchain-api & The-Factory
  - drop `blockchain-api`


JackBekket:
Looking pretty elegant 🤘

JackBekket:
Looking pretty elegant 🤘

JackBekket:
I will wait till you move tests from here - https://github.com/sonm-io/blockchain-api/tree/master/tests

Please, notice me once, when it will be done

JackBekket:
I will wait till you move tests from [blockchain-api](https://github.com/sonm-io/blockchain-api/tree/master/tests)

Please, notice me once, when it will be done

Sokel:
@JackBekket we will not transfer this examples as is.
For ours enviroment by condition, we need embeded unit auto tests (*_test.go).
These codes will be transfered by if needed.

Sokel:
@jorpic may you separate *.sol in their own directory?

Sokel:
@jorpic and there...
will you wrap instruction in Makefile in their own blocks and include it to other?

Sokel:
> requires truffle to be installed on developers machine

its works correctly with CI (as travis)?

> no need to store generated code in repo

 it haves broken path to include any packages

JackBekket:
Agreed about autotests, but also I strongly needed to have at least one scanario test, which include normal cycle work - create wallets, register in whitelist, token pay proceeded. 

I belive it can be found [here](https://github.com/sonm-io/blockchain-api/blob/master/tests/test.go)

Sokel:
@JackBekket it must will written after package structure.
you can suggest an algorithm to be implemented in example you need

jorpic:
**Updates**

  - `Makefile` is updated to compile `contracts/*.sol` and `contracts/api.go`
  - `.travis.yml` is updated to run `npm install` before building with `make`
  - global truffle installation is not required, the one from `devDependencies` is used

**TODO**

  - refactor contracts
  - write some tests for `contracts/api.go`
     - should be possible to run them with clean `testrpc`
     - tests in golang style

Makefile
```diff
@@ -41,7 +41,7 @@ build_contracts:
 	${GO} generate ./contracts
 	${GO} build ./contracts/api.go
 
-build: build_bootnode build_hub build_miner build_cli build_contracts
+build: build_contracts build_bootnode build_hub build_miner build_cli
```
Sokel: contracts api must be compiled before, because it would be used at miner and hub

Makefile
```diff
@@ -34,10 +35,14 @@ build_cli:
 	@echo "+ $@"
 	${GO} build -o ${CLI} ${GOCMD}/cli
 
+build_contracts:
+	@echo "+ $@"
+	cd ./blockchain && npm install && ${TRUFFLE} compile
```
Sokel: yeap, its aimed to blank installation, in other case - npm return true after check

Makefile
jorpic: I think we need to use `make`s ability to track dependencies and and ordering build steps.
This will also lead to faster rebuilds.

PR #14, opened 2017-07-29
Collect and provide runtime Miner metrics

### LINKED ISSUE #5: Metrics collection from miner (CPU, Memory)
Collect metrics collection from miner (CPU, Memory), report to hub.
- [x] Metrics from Docker (used CPU, Memory).
- [ ] Metrics from OS (total CPU, Memory).
- [x] Communication through gRPC.
- [x] Hub requests report.
- [x] Miner provide report.

3Hren:
This PR activates runtime Docker statistics collection from all running containers. That statistics can be obtained using gRPC miner server.

There is currently unresolved question of what to do with metrics from quickly-dying containers. It seems that we'll need some kind of purgatory with these dead containers metrics that is cleared on first info request. Then it is required that all info requests must be strongly authenticated to avoid metrics lose.

Additionally a CLI integration has been made, an output example:

```bash
esafronov@esafronov:~/go/src/github.com/sonm-io/core|feat/runtime-metrics
⇒  ./sonmcli --hub=localhost:10001 info --miner="[::1]:55796"
{}

esafronov@esafronov:~/go/src/github.com/sonm-io/core|feat/runtime-metrics⚡
./sonmcli --hub=localhost:10001 info --miner="[::1]:55531"
{"Stats":{"2f74767a0b053e3f2049307d7b7b0e19c64d18a9858e737b11c71dbd9485756c":{"CPU":{"totalUsage":145960673},"Memory":{"maxUsage":2936832}}}}

esafronov@esafronov:~/go/src/github.com/sonm-io/core|feat/runtime-metrics⚡
./sonmcli --hub=localhost:10001 info --miner="[::1]:55531"
{"Stats":{"2f74767a0b053e3f2049307d7b7b0e19c64d18a9858e737b11c71dbd9485756c":{"CPU":{"totalUsage":152393152},"Memory":{"maxUsage":2936832}}}}

esafronov@esafronov:~/go/src/github.com/sonm-io/core|feat/runtime-metrics⚡
./sonmcli --hub=localhost:10001 info --miner="[::1]:55531"
{"Stats":{"2f74767a0b053e3f2049307d7b7b0e19c64d18a9858e737b11c71dbd9485756c":{"CPU":{"totalUsage":152393152},"Memory":{"maxUsage":2936832}}}}
```

Found inconsistency and inconvenience: 
- Seems like generated proto and gRPC files should not be a part of commit, instead we should generate them during our build process.
- There are some inconvenience that makes hard to reuse proto files with import etc. That led to minor copy/paste, but we should also eliminate it.

Partially closes #5, because there are total resources given from an OS is left to be collected.


insonmnia/miner/overseer.go
```diff
@@ -173,6 +201,10 @@ func (o *overseer) collectStats() {
 			}
 			o.mu.Unlock()
 			for _, id := range ids {
+				if len(id) == 0 {
```
noxiouz: How is that possible to have an empty id?

3Hren: Because of pool. Do we ever need it?

noxiouz: stringArrayPool.Put(ids[:0]) it shrinks the slice, but saves capacity

insonmnia/hub/server.go
```diff
@@ -68,6 +68,40 @@ func (h *Hub) List(context.Context, *pb.ListRequest) (*pb.ListReply, error) {
...
+	if !ok {
+		return nil, status.Errorf(codes.NotFound, "no such miner")
+	}
+
+	resp, err := client.Client.Info(ctx, &pbminer.InfoRequest{})
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to fetch info: %v", err)
+	}
+
+	// TODO: Reuse proto files with imports. Currently it's problematic, because of some reasons.
```
antmat: See https://github.com/sonm-io/core/pull/17/files#diff-b67911656ef5d18c4ae36cb6741b7965R62

insonmnia/miner/overseer.go
3Hren: Yeah, after the first collecting. But *while* the first collecting there are empty strings.

PR #16, opened 2017-07-30
WIP: Feature/hub config

sshaman1101:
**Review required** 
Adding yaml config, adding cmd flag for config path;

Please say, is daemon must crash if there no config file or config file is invalid? Or just use default values instead?


3Hren:
Hmm, something went wrong with our issue-assignment intercooperativity. According to the release plan after brainstorming I was pretty sure, that this task was assigned to me, but whatever :). Summoning @iolebedev (JFYI).

First of all, I'm sure that the config must be hierarchical instead of plain. It improves readability, makes it structured and allows to reuse components in the future. Secondary it must be an interface to be able to mock where required.

sshaman1101:
I was ask Anton to assign me to this task :)
+1 for config interface, also may be reused for another packages.

What about config defaults? I was discuss it with Anton, he suggest to apply some default values if there is no config file or config is filled improperly.
What is you opinion on this?

3Hren:
IMO excessive defaults usage is evil, because it's some kind of blackbox magic. Everyone's forgot their real values after some period of time and the only way to remember them (if they aren't obvious) is to look into the code.

Moreover there will be a time when we can't ignore defaults - config backward compatibility, when adding new entities. Until then I'd rather avoid defaults.

Instead we can provide a clear way to detect misconfiguration location and better error messages. This is also must be covered with tests, as it's UX.

sshaman1101:
> Instead we can provide a clear way to detect misconfiguration location and better error messages. 

I am trying to do it [here](https://github.com/sonm-io/core/blob/feature/hub-config/insonmnia/hub/config.go#L17), but this validation method may be so shitty.
Maybe it will be better to pick up some configuration library that able to perform validation via structure tags (like [configor](https://github.com/jinzhu/configor))?




> First of all, I'm sure that the config must be hierarchical instead of plain. It improves readability, makes it structured and allows to reuse components in the future

do you mean this?
```yaml
hub:
  grpc_endpoint: ":10001"
  miner_endpoint: ":10002"
``` 

instead of
```yaml
grpc_endpoint: ":10001"
miner_endpoint: ":10002"
```

cmd/hub/main.go
```diff
@@ -12,9 +13,20 @@ import (
...
+var (
+	configPath = flag.String("config", "hub.yaml", "Path to hub config file")
+)
+
 func main() {
+	flag.Parse()
+
+	conf, err := hub.NewConfig(*configPath)
+	if err != nil {
+		panic("Cannot load config file: " + err.Error())
```
noxiouz: no panic please :)

insonmnia/hub/config.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+type HubConfig struct {
+	GRPCEndpoint  string   `yaml:"grpc_endpoint"`
+	MinerEndpoint string   `yaml:"miner_endpoint"`
+	Bootnodes     []string `yaml:"bootnodes"`
+}
+
+func (conf *HubConfig) validate() error {
+	var errs []string
```
noxiouz: Look at this https://github.com/hashicorp/go-multierror
May be it fits better

cmd/hub/main.go
sshaman1101: Should I use os.Exit(1) instead?

noxiouz: Do not forget to log error

insonmnia/hub/config.go
```diff
@@ -0,0 +1,57 @@
...
+	"io/ioutil"
+	"strings"
+
+	"gopkg.in/yaml.v2"
+)
+
+type HubConfig struct {
+	GRPCEndpoint  string   `yaml:"grpc_endpoint"`
+	MinerEndpoint string   `yaml:"miner_endpoint"`
+	Bootnodes     []string `yaml:"bootnodes"`
```
noxiouz: If Bootnodes has special format, it might be better to define a custom type with Unmarshaller

Sokel: In used libs from go-ethereum, enode's addreses casts from rawurl(string) to *Node object.
It's doesn't have ready-to-use pattern now.

3Hren:
> do you mean this?

Some kind of.

cmd/hub/main.go
3Hren: It should be not 0. However some OS specifies their own exit codes for a good style. For example FreeBSD recommends to exit with EX_USAGE status.

For this moment I'd rather exit with 1 for the sake of not make things complex.

insonmnia/hub/config.go
3Hren: Why not to just break with the first found error?

insonmnia/hub/server.go
```diff
@@ -18,16 +18,14 @@ import (
...
-	// TODO: make it configurable
-	externalGRPCEndpoint         = ":10001"
-	minerHubInterconnectEndpoint = ":10002"
-)
+
 
 // Hub collects miners, send them orders to spawn containers, etc.
 type Hub struct {
 	ctx           context.Context
+	grpcEndpoint  string
```
3Hren: IP + port instead?

noxiouz: grpcDialer accepts target as string, because target is kind of service name (like in Cociane)

3Hren: Then convert it again when it's time for.

cmd/hub/main.go
```diff
@@ -12,9 +13,21 @@ import (
...
+)
+
 func main() {
+	flag.Parse()
 	ctx := context.Background()
-	h, err := hub.New(ctx)
+
+	conf, err := hub.NewConfig(*configPath)
+	if err != nil {
+		ctxlog.GetLogger(ctx).Fatal("Cannot load config", zap.Error(err))
```
3Hren: Fatal calls os.Exit(1) [for you](https://godoc.org/go.uber.org/zap#Logger.Fatal).

insonmnia/hub/config_test.go
```diff
@@ -0,0 +1,67 @@
...
+  grpc_endpoint: ""
+  miner_endpoint: ":10002"`
+
+	err := createTestConfigFile(raw)
+	assert.Nil(t,err)
+
+	conf, err := NewConfig(testHubConfigPath)
+	assert.Nil(t, conf)
+	assert.Contains(t, err.Error(), "GRPCEndpoint is required")
+}
```
3Hren: Newline please.

cmd/hub/main.go
sshaman1101: Hmm, not so explicit behaviour. Maybe rewrite using .Error() and then explicitly call os.Exit(1)?

3Hren: Agree.

insonmnia/hub/server.go
noxiouz: ok

insonmnia/hub/server.go
```diff
@@ -18,16 +18,14 @@ import (
...
 	externalGrpc  *grpc.Server
+	minerEndpoint string
```
noxiouz: so do it the same way as grpcEndpoint

sshaman1101:
Pls do not merge this now, I think i should add same feature to miner too. Also maybe it will be useful to move `insonmnia/hub/logger.go` into separated package and make buildLogger exportable.

Issue #6, opened 2017-07-27
Miner set restrictions on resource pool


iolebedev:
Miner set restrictions on resource pool
- Launch containers using parent cgroups
- Setup parent cgroup, where mounted, permissions
- Rebuild cgroup on restart
- Report resources on handshake

noxiouz:
closed via #18

PR #18, opened 2017-07-30
resource pool

noxiouz:
Related to #6 

Blocked by #16 (needs configuration)

### DIFF
--- a/insonmnia/miner/cgroup_linux.go
+++ b/insonmnia/miner/cgroup_linux.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"github.com/containerd/cgroups"
+	"github.com/mitchellh/mapstructure"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 )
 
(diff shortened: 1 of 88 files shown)


insonmnia/miner/server.go
```diff
@@ -389,6 +394,19 @@ func New(ctx context.Context, config *MinerConfig) (*Miner, error) {
 		cancel()
 		return nil, err
 	}
+
+	var deleter cGroupDeleter
+	if platformSupportCGroups {
```
antmat: This can be simplified if we use NilDeleter

insonmnia/miner/cgroup_nonlinux.go
```diff
@@ -0,0 +1,19 @@
...
+
+type Resources interface{}
+
+const (
+	platformSupportCGroups = false
+	parentCgroup           = ""
+)
+
+func initializeControlGroup(*Resources) (cGroupDeleter, error) {
+	return nil, fmt.Errorf("%s does not support Control Groups", runtime.GOOS)
```
antmat: How about some kind of NilDeleter instead of error? Looks like it is not very usefull. And we already have a flag indicating cgroups support

insonmnia/miner/cgroup_linux.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+	default:
+		return nil, fmt.Errorf("failed to load parent cgroup %s: %v", parentCgroup, err)
+	}
+
+	if res == nil {
+		return control, nil
+	}
+
+	if control.Update((*specs.LinuxResources)(res)); err != nil {
```
antmat: err assignement missed

noxiouz: nice catch

insonmnia/miner/cgroup_linux_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseResources(t *testing.T) {
+	assert := assert.New(t)
+	defer deleteTestConfigFile()
+	raw := `
+miner:
+  hub_address: 127.0.0.1
```
3Hren: This can be omitted.

insonmnia/miner/config.go
```diff
@@ -4,17 +4,21 @@ import (
 	"github.com/jinzhu/configor"
 )
 
-type MinerConfig struct {
```
3Hren: Changes in this file are mostly out of the scope (and is done in https://github.com/sonm-io/core/pull/40), except for adding limitations option.

Also may be rename to `restrictions`?

insonmnia/miner/server.go
```diff
@@ -160,7 +159,7 @@ func (m *Miner) setStatus(status *pb.TaskStatus, id string) {
 func (m *Miner) listenForStatus(statusListener chan pb.TaskStatus_Status, id string) {
 	select {
 	case newStatus := <-statusListener:
-		m.setStatus(&pb.TaskStatus{newStatus}, id)
+		m.setStatus(&pb.TaskStatus{Status: newStatus}, id)
```
3Hren: I'd rather split these changes into a separate PR.

insonmnia/miner/config.go
noxiouz: I prefer the same naming as it's in Docker.

3Hren: Got it.

Issue #34, opened 2017-08-05
Allow JSON output in CLI


3Hren:
This is very helpful for monitoring robots, as they can easily parse JSON while having problems with the current human-readable output.

Possibly using optional argument `--output=[plain(default), json]`

cc @sshaman1101.

sshaman1101:
Good idea, I will implement this.
Maybe it will be better to add config file? like `~/.sonm.yaml` and keep hub address and also output mode?

PR #35, opened 2017-08-05
feat: nice RAM output

3Hren:
This PR fixes fraction truncating, while displaying total RAM consumed by a task. Also it automatically changes units to properly show memory consumption in bytes, kilobytes, megabytes etc.

However in JSON output (#34) it is better to display in bytes.

cc @sshaman1101


noxiouz:
+1

PR #39, opened 2017-08-06
fix: create miner logger again

3Hren:
This commit fixes accidentally removed logger initialization in miner. For consistency with a Hub it is created in the main and assigned to the context.

Logger is not a part of a Hub construction, so this decomposition is more proper.


sshaman1101:
And we must do the same with Hub, isn't is?

PR #43, opened 2017-08-06
Refact/pb single package

antmat:
Now all proto files belong to one package. Also source code is being built on the fly to reduce number of conflicts in generated code.
Still not sure if this approach looks better than previous one, so feel free to discuss.
Chained with https://github.com/sonm-io/core/pull/42
@noxiouz @3Hren PTAL


noxiouz:
it requires grpc-go and protobuf

3Hren:
Rebase.

Issue #45, opened 2017-08-07
Miner sometimes hangs during discovery start up


3Hren:
Here: https://github.com/sonm-io/core/blob/master/insonmnia/miner/server.go#L388.

Sokel:
@3Hren may you set log level for go-ethereum to 10, [this](https://github.com/sonm-io/core/blob/master/common/const.go#L10), and give it me?

3Hren:
It printed nothing:

```
⇒  ./sonmminer
2017-08-07T13:24:34.439+0300	INFO	miner/overseer.go:141	subscribe to Docker events	{"since": "1502101474"}
INFO [08-07|13:24:34] Starting P2P networking
```

I've checked the Hub with this logging level and it emitted tons of logs.

3Hren:
@Sokel is this resolved after eliminating whisper?

Sokel:
@3Hren yeap

PR #46, opened 2017-08-07
fix: use discovery on demand

3Hren:
This fixes meaningless hanging when unable to discover hubs.

cc @Sokel @noxiouz


Sokel:
related to #45.

Sokel:
its already might be as hotfix, i have any plans for this in closely

PR #47, opened 2017-08-07
fix: use proper context for logging RPC

3Hren:
GRPC provides its own detached context, so using it emits default logging configuration.

This commit fixes that behavior by using our context with preconfigured logging.


noxiouz:
The problem is that it wiould use global logger instead of attached to a context, because grpc context has no attached logger. The better approach is to implement interceptors. 
Like this https://github.com/grpc-ecosystem/go-grpc-middleware/tree/master/logging/zap

3Hren:
Not so global, but our's hub attached, which is configured by us.

PR #48, opened 2017-08-07
Feature/add contracts with go tests

jorpic:
This adds few contracts form [The-Factory](https://github.com/sonm-io/The-Factory).
The goal is to have consistent builds and to run smart contract aware tests on each build.

Features:

  - Single `npm install` command to install all smart contract related tools and dependencies.
  - `go generate` is used to generate contract wrappers on each build
    - no need to install `truffle` on dev's machine
    - no need to install `abigen` on dev's machine as it is imported from go-ethereum by [contracts/generate_api.go](https://github.com/sonm-io/core/blob/b720b78d00010d2d23a80582d68be0232a94da2f/contracts/generate_api.go)
  - `make test` runs tests with `testrpc`
    - real automatic and reproducible tests, no need to setup `geth` on dev's machine
    - testing smart contracts in golang


Sokel:
1) `make all` instruction doesn't work on blank installation
```
./node_modules/truffle/build/cli.bundled.js compile
make: ./node_modules/truffle/build/cli.bundled.js: No such file or directory
make: *** [build_contracts] Error 1
```
need to place `npm install` in one of instructions

2) `contracts` is not obvious naming above blockchain integration to core
its may be `blockchain` package name

3) truffle may be replaced with solc and composes compilation instructions into abi-generator

4) all files in one directory.

 You have in one:
- solidity contracts source
- tests for contracts
- helpers for contracts test
- go generator for api
- api
- test for api (in future?)
It may be separated into different directories?

5) need to include `test_contacts` instruction to `test`

6) more: may you research [solcover](https://github.com/JoinColony/solcover) util for this installation?

7) generated files in tracking & broken import, in any suggest we may to use `ignore, tracker by git` - what you think about this suggest?

jorpic:
> need to place `npm install` in one of instructions

Added it to `build` task, but I think this is quite a questionable decision:

  - `npm install` is slow and requires internet connection even if everything is installed
  - someone may prefer to use `yarn` instead of `npm`

----

> contracts is not obvious naming above blockchain integration to core

Done. All contract-related code is moved to `blockchain` directory.

---

> truffle may be replaced with solc and composes compilation instructions into abi-generator

Truffle simplifies managing Solidity dependencies and generates easy-to-use artifacts.

---

> all files in one directory.

Fixed. They are sorted into `contracts` / `tests` / `utils` directories now.

---

> need to include test_contacts instruction to test

Done.

---

> more: may you research solcover util for this installation?

I think this is a great idea for a new issue and another pull-request.

--- 
> generated files in tracking & broken import, in any suggest we may to use ignore, tracker by git

Generated contract wrappers are under version control now.

Sokel:
LGTM

PR #52, opened 2017-08-07
feat: handshake

### LINKED ISSUE #29: Emit miner's OS limits with the handshake

3Hren:
This PR allows a Miner to provide its total resources available to a Hub that is connected to it.

For now we assume that only one Hub can be connected to a Miner. If not so - we should control taken resources in Miner itself and notify all hubs with changes.

Consider it as a first iteration forwarding to smart scheduling.

Closes #29.


insonmnia/hub/miner_ctx.go
```diff
@@ -70,9 +82,35 @@ func createMinerCtx(ctx context.Context, conn net.Conn) (*MinerCtx, error) {
...
+	if err := m.handshake(); err != nil {
+		m.Close()
+		return nil, err
+	}
+
 	return &m, nil
 }
 
+func (m *MinerCtx) handshake() error {
+	log.G(m.ctx).Info("sending handshake to a Miner", zap.Any("addr", m.conn.RemoteAddr()))
```
noxiouz: zap.Stringer

PR #59, opened 2017-08-08
fix: MinerBuilder mandatory param + pass context in main

antmat:
Config is mandatory for builder, also we need to pass param in main when creating server

### DIFF
--- a/insonmnia/miner/builder.go
+++ b/insonmnia/miner/builder.go
@@ -7,6 +7,7 @@ import (
 
 	log "github.com/noxiouz/zapctx/ctxlog"
 	"github.com/pborman/uuid"
+	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


3Hren:
If a config is required it is better to construct builder itself using some kind of `NewMinerBuilder(cfg) -> Builder`.

insonmnia/miner/builder.go
```diff
@@ -42,22 +43,26 @@ func (b *MinerBuilder) Overseer(ovs Overseer) {
...
 
 	if b.collector == nil {
 		b.collector = resource.New()
 	}
 
-	if b.ctx == nil {
-		b.ctx = context.Background()
+	if b.ip == nil {
+		addr, err := util.GetPublicIP()
+		if err != nil {
```
noxiouz: In all retun err branches it's better to have cancel()

PR #62, opened 2017-08-11
feature: task description via yaml file

sshaman1101:
This PR allows to pass task description via yaml


sshaman1101:
This PR allows to set SSH key to get access to container with task.
But it may be bad solution because many users will have one and only one ssh key on their system.
I suggest to move ssh-pub-key setting to cli config itself (`~/.sonm/cli.yaml`)
@antmat @noxiouz PTAL

PR #63, opened 2017-08-11
feat: ssh embedded server for container access

antmat:
Early version without authentication.
@3Hren @noxiouz @Sokel @sshaman1101 
PTAL

### DIFF
--- a/insonmnia/miner/builder.go
+++ b/insonmnia/miner/builder.go
@@ -56,8 +56,9 @@ func (b *MinerBuilder) UUID(uuid string) *MinerBuilder {
 	return b
 }
 
-func (b *MinerBuilder) SSH(ssh SSH) {
+func (b *MinerBuilder) SSH(ssh SSH) *MinerBuilder {
 	b.ssh = ssh
+	return b
 }
 
 func (b *MinerBuilder) Build() (miner *Miner, err error) {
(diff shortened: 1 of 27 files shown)


Sokel:
good;
shh will be always on?

antmat:
We can make it configurable, if we decide to.

Sokel:
I think, it will be configurable by flag -noshh or any

Sokel:
About authentication: would you research it with *ecdsa.Private key?

noxiouz:
@Sokel https://godoc.org/golang.org/x/crypto/ssh#NewPublicKey
It should work out of the box

insonmnia/miner/builder.go
```diff
@@ -53,6 +54,10 @@ func (b *MinerBuilder) UUID(uuid string) *MinerBuilder {
 	return b
 }
 
+func (b *MinerBuilder) SSH(ssh SSH) {
+	b.ssh = ssh
```
noxiouz: return *MinerBuilder?

insonmnia/miner/ssh.go
```diff
@@ -0,0 +1,106 @@
...
+		outputErr <- err
+	}()
+
+	go func() {
+		defer stream.CloseWrite()
+		_, err := io.Copy(stream.Conn, session)
+		outputErr <- err
+	}()
+
+	err = <-outputErr
```
noxiouz: goroutine leak: you write 2 errors in the chan, but read only 1

Issue #49, opened 2017-08-07
Authenticate Hub/Miner via WhiteList smart contract


jorpic:
Hub and miner should be able to:
 - authenticate each other using ETH-addresses and ECDSA;
 - get some info (e.g. deposit, status) about each other from the WhiteList smart contract.

Features:

 - [ ] blockchain-related configuration section in config file
  - Eth JSON-RPC provider to use
  - Eth network id
  - Eth credentials (private key / path to json wallet / etc)
 - [ ] allow to run hub/miner in *trust all* mode, i.e. without blockchain-related configuration
 - [ ] basic challenge-response protocol to authenticate each other
 - [ ] get peer info from `WhiteList` contract
   - there is no use for this info currently, so it is enough to drop it to log

jorpic:
Hey guys, please check this authentication protocol proposal (it is based on the protocol by Denning&Sacco which is a modification of the classic Needham&Schroeder).

Goals of the protocol:

  - exchange identities (ETH-addresses)
  - prove that contragents are aware whom they are communicating with
  - generate private session id tat can be used to establish secure communication channel
  - some messages can be used as a proof of communication in ETH smart-contract

Protocol steps:

  - **A sends** `AuthInitRequest`
    - `timestamp`
    - nonce hash: `nonceHash = SHA3(nonceA | timestamp)`
    - message signature
  - **B checks**
    - signature is valid, get `pkA` (A's public key), `addrA` (A's ETH address)
    - `timestamp` is valid (within 30 minutes interval)
  - **B sends** `AuthInitResponse` encrypted with `pkA`
    - `addrA`
    - `nonceHash`
    - `nonceB`
    - message signature
  - **A checks**
    - signature is valid, get `pkB` (B's public key), `addrB` (B's ETH address)
    - `nonceHash` is valid
  - **A sends** `AuthCommitRequest` encrypted with `pkB`
   - `addrB`
    - `nonceA`, `nonceB`
    - message signature
  - **B checks** `
    - signature is valid
    - `nonceHash = SHA3(nonceA | timestamp)`
  - **B sends** `AuthCommitResponse` encrypted with `pkA`
   - `addrA`
    - `nonceA`, `nonceB`
    - message signature

Communication is complete, `SHA3(nonceA | nonceB)` can be used as private session id.

Notes:
  - each message contains `message type` field
  - secp256k1 is used to sign messages
  - ECIES is used to encrypt messages

Sokel:
[...]

PR #70, opened 2017-08-15
Share hub's pubkey on discovery

jorpic:
This is a baby first step to authenticating hub/miner via white list in ETH blockchain:

 - hub gets his private key from config
 - hub signs his address announcement with his private key
 - miner gets hub's public key from address announcement message

Now miner is ready to check white list in order to get hub's deposit info and rating.


Sokel:
#49

PR #72, opened 2017-08-15
ssh: log ssh.Run error

noxiouz:


insonmnia/miner/server.go
```diff
@@ -453,7 +453,9 @@ func (m *Miner) Serve() error {
 		go func() {
 			log.G(m.ctx).Info("starting ssh server")
 			defer wg.Done()
-			m.ssh.Run(m)
+			if sshErr := m.ssh.Run(m); sshErr != nil {
+				log.G(m.ctx).Error("failed to run SSH server", zap.Error(sshErr))
```
antmat: This line would be present also on normal shutdown

noxiouz: done

PR #73, opened 2017-08-15
feat: passthrough for task logs

antmat:
Support to deliver task logs from miner through hub to client. @noxiouz @3Hren @Sokel  @sshaman1101 PTAL


noxiouz:
💩

cmd/cli/commands/tasks.go
```diff
@@ -98,6 +106,61 @@ var taskListCmd = &cobra.Command{
...
+	Use:     "logs <task_id>",
+	Short:   "Show task status",
+	PreRunE: checkHubAddressIsSet,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) < 1 {
+			return errTaskIDRequired
+		}
+		taskID := args[0]
+		req := pb.TaskLogsRequest{}
+		if logType == "stderr" {
```
noxiouz: switch logType {
case "stderr": 
}...

cmd/cli/commands/tasks.go
```diff
@@ -98,6 +106,61 @@ var taskListCmd = &cobra.Command{
...
+var taskLogsCmd = &cobra.Command{
+	Use:     "logs <task_id>",
+	Short:   "Show task status",
+	PreRunE: checkHubAddressIsSet,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) < 1 {
+			return errTaskIDRequired
+		}
+		taskID := args[0]
+		req := pb.TaskLogsRequest{}
```
noxiouz: set fields here instead of line 129 ...

proto/insonmnia.proto
```diff
@@ -60,3 +60,22 @@ message ContainerRestartPolicy {
...
+    string id = 2;
+    string since = 3;
+    bool addTimestamps = 4;
+    bool Follow = 5;
+    string Tail = 6;
+    bool Details = 7;
+}
+
+message TaskLogsChunk {
+    string data = 2;
```
noxiouz: bytes? to avoid casting from []byte to string

insonmnia/miner/server.go
```diff
@@ -371,6 +373,40 @@ func (m *Miner) sendUpdatesOnRequest(server pb.Miner_TasksStatusServer) {
...
+	}
+	buffer := make([]byte, 100*1024)
+	for {
+		readCnt, err := reader.Read(buffer)
+		if readCnt != 0 {
+			server.Send(&pb.TaskLogsChunk{string(buffer[:readCnt])})
+		}
+		if err == io.EOF {
+			return nil
+		} else if err != nil {
```
noxiouz: if err != nil is sacred in go! It's better to do same across the code. Look at hub/server.go +210

insonmnia/miner/server.go
```diff
@@ -371,6 +373,40 @@ func (m *Miner) sendUpdatesOnRequest(server pb.Miner_TasksStatusServer) {
...
+		Since:      request.Since,
+		Timestamps: request.AddTimestamps,
+		Follow:     request.Follow,
+		Tail:       request.Tail,
+		Details:    request.Details,
+	}
+	reader, err := m.ovs.Logs(server.Context(), cid, opts)
+	if err != nil {
+		return err
+	}
```
noxiouz: defer reader.Close()

PR #90, opened 2017-08-18
feat: NAT type detection

3Hren:
This commit activates automatic NAT discovery using STUN protocol during miner startup. This is planned to be used to bypass NAT restrictions to make miners available from the private networks.

Unfortunately, NAT detection may take a long. It's about 20-30 seconds taken to check all available NAT types, but it's pretty precise.

### DIFF
--- a/insonmnia/miner/builder.go
+++ b/insonmnia/miner/builder.go
@@ -5,13 +5,13 @@ import (
 
 	"net"
 
+	"github.com/ccding/go-stun/stun"
 	log "github.com/noxiouz/zapctx/ctxlog"
 	"github.com/pborman/uuid"
 	"github.com/pkg/errors"
 	"github.com/sonm-io/core/insonmnia/hardware"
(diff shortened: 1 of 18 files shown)


insonmnia/miner/builder.go
```diff
@@ -75,11 +76,20 @@ func (b *MinerBuilder) Build() (miner *Miner, err error) {
 	}
 
 	if b.ip == nil {
-		addr, err := util.GetPublicIP()
+		log.G(b.ctx).Debug("discovering public IP address with NAT type, this may take a long ...")
+
+		client := stun.NewClient()
```
antmat: Does it use some default STUN server? Maybe we need to make it also configurable

3Hren: In fact it doesn't help much, but we'll come to this soon or later. Just haven't came up with the appropriate name.

3Hren: Anyway fixed in #92.

PR #93, opened 2017-08-21
feat: smart routing via IPVS

### LINKED ISSUE #91: NAT bypassing
After struggling for a couple of days I've came to the idea how we can allow miners without public IPs to be able to work.

I've made a prototype and it works! Multiple PRs incoming ...

Le plan:

- [x] Detect NAT type on the miner.
  - If there is no NAT type or it's weak like symmetric UDP firewall - that's fine. Allow them to publish themselves on their white IPs.
  - If there is NAT or some other firewall which restricts public access - then we're in trouble.
- [x] Provide this information to the Hub.
- [x] Hub can organise virtual private network with such miners, allowing requests to be reachable at least from the Hub's endpoint. Either directly or using NAT piercing strategy with further ipip tunnelling.
- [x] For miners with public IPs there will be direct access with containers. That information is provided using **start** and **info** requests.
- [x] For miners that can't be reachable from the WAN we can provide dynamic routing using IPVS configured as NAT. Then provide our own public IP and remapped port for virtual services as an endpoint(s) to the task.

Sudden profits:
- Smart and configurable (rr, lwc, wrr, etc.) load balancing between similar containers from the same task. For now we can have only 1 container per task, however it is possible to perform auto-scaling.
- 1:1 port mapping is also becomes possible, especially for applications that can't be configured.
- Both TCP/UDP routing.

Pros:
- Fast, as it is L3 forwarding and is done in the kernel.
- Fully optional. Some kind as extension for miners that can't obtain public IP for some reasons.

Cons:
- Miner should set the Hub as a default gateway (or precisely for a given restricted port range using iptables), otherwise packets will be dropped by the kernel as unexpected, even if the client is reachable directly from miners.
- Routed traffic must be additionally accounted (however it's possible to). Not sure it's cons :).
- Increased timings (not much), as it's an additional hop between clients and miners.
- Hubs are restricted to be run on linux machine, because of netlink and lvs.

3Hren:
This PR introduces traffic routing for Miners under the firewall using IPVS.

TBD:
- [x] Route Docker network properly.
- [x] Dismiss virtual services on task stop.
- [x] Dismiss virtual services on shutdown.
- [x] Return properly remapped endpoints on task start reply.

Closes #91.

### DIFF
--- a/insonmnia/gateway/gateway_linux.go
+++ b/insonmnia/gateway/gateway_linux.go
@@ -20,6 +20,7 @@ var (
 	ErrIPVSFailed      = errors.New("error while calling into IPVS")
 	ErrBackendExists   = errors.New("backend already exists")
 	ErrServiceNotFound = errors.New("virtual service not found")
+	ErrBackendNotFound = errors.New("backend not found")
 )
 
 type service struct {
@@ -65,6 +66,10 @@ func NewGateway(ctx context.Context) (*Gateway, error) {
 	return gateway, nil
 }
 
+// TODO (3Hren): func (g *Gateway) GetServices() ([]string, error).
+// TODO (3Hren): func (g *Gateway) GetBackends(vsID string) ([]string, error).
+// TODO (3Hren): func (g *Gateway) GetBackend(vsID, rsID string) (*RealOptions, error).
+
 // CreateService registers a new virtual service with IPVS.
 func (g *Gateway) CreateService(vsID string, options *ServiceOptions) error {
 	g.mu.Lock()
@@ -73,6 +78,7 @@ func (g *Gateway) CreateService(vsID string, options *ServiceOptions) error {
 	return g.createService(vsID, options)
 }
 
+// CreateBackend registers a new backend with the virtual service.
 func (g *Gateway) CreateBackend(vsID, rsID string, options *RealOptions) error {
 	g.mu.Lock()
 	defer g.mu.Unlock()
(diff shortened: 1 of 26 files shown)


insonmnia/gateway/pool.go
```diff
@@ -0,0 +1,60 @@
+package gateway
+
+import (
+	"github.com/pkg/errors"
```
noxiouz: just errors, I suppose

insonmnia/miner/marshal.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/ccding/go-stun/stun"
+
+	pb "github.com/sonm-io/core/proto"
+)
+
+func marshalNATType(nat stun.NATType) pb.NATType {
+	switch nat {
+	case stun.NATNone:
+		return pb.NATType_NONE
+	case stun.NATError:
```
noxiouz: case stun.NATError, stun.NATUnknown:
....

insonmnia/gateway/pool.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return port.(uint16), nil
+}
+
+func (p *PortPool) Retain(ID string) error {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if port, exists := p.used[ID]; exists {
+		p.queue.Enqueue(port)
```
noxiouz: !exists return error

p..Enqueue
delete
...

3Hren: We need a port to enqueue and to check in a single map touching.

3Hren:
That's all. Discuss!

insonmnia/gateway/gateway_linux.go
```diff
@@ -0,0 +1,243 @@
...
+		zap.String("service", vsID),
+		zap.String("backend", rsID),
+		zap.String("host", host),
+		zap.Uint16("port", port),
+	)
+
+	if err := g.ipvs.DelDestPort(host, port, backendHost, backendPort, protocol); err != nil {
+		log.G(g.ctx).Error("failed to remove backend",
+			zap.String("service", vsID),
+			zap.String("backend", rsID),
```
noxiouz: zap.Error(err)

PR #102, opened 2017-08-22
feat: make gateway optional

3Hren:
This commit allows to disable Hub as a gateway. Intended for testing environments and not to break brains configuring linux internals to make things work.

### DIFF
--- a/insonmnia/hub/server.go
+++ b/insonmnia/hub/server.go
@@ -296,9 +296,21 @@ func New(ctx context.Context, cfg *HubConfig) (*Hub, error) {
 		return nil, errors.Wrap(err, "malformed ethereum private key")
 	}
 
-	gate, err := gateway.NewGateway(ctx)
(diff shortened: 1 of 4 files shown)


insonmnia/hub/server.go
```diff
@@ -296,17 +296,29 @@ func New(ctx context.Context, cfg *HubConfig) (*Hub, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		if len(cfg.GatewayConfig.Ports) != 2 {
+			return nil, errors.New("gateway ports must be a range of two values")
+		}
+
+		portRangeFrom := cfg.GatewayConfig.Ports[0]
+		portRangeSize := cfg.GatewayConfig.Ports[1] - portRangeFrom
```
noxiouz: I would check that Ports[1] > Ports[0]. Just in case :)

PR #106, opened 2017-08-24
feat: virtual services statistics

3Hren:
This commit allows to collect virtual services statistics for further traffic accouting.

All requests that are redirected through the gateway consumes network, which is the limited resource for us. Collecting bytes transmitted and received as like as the number of connections consumed allows to properly calculate how much a specific miner affects our network.

There are separate metrics for a specific miner, allowing us to gain statistics for each miner connected (still cleared after disconnection).

Dead virtual services does not clear their accounted traffic consumed.

### DIFF
--- a/insonmnia/hub/router_linux.go
+++ b/insonmnia/hub/router_linux.go
@@ -12,7 +12,7 @@ type ipvsRouter struct {
 	pool     *gateway.PortPool
 	minerID  string
 	metrics  map[string]*gateway.Metrics
-	services map[string]bool
+	services map[string]struct{}
 	mu       sync.Mutex
 }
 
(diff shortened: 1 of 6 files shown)


insonmnia/hub/router_linux.go
```diff
@@ -2,22 +2,27 @@ package hub
...
 type ipvsRouter struct {
-	gateway *gateway.Gateway
-	pool    *gateway.PortPool
-	ids     map[string]bool
-	mu      sync.Mutex
+	gateway  *gateway.Gateway
+	pool     *gateway.PortPool
+	minerID  string
+	metrics  map[string]*gateway.Metrics
+	services map[string]bool
```
noxiouz: if it emulates set, it's better to use to map[T]struct{}

PR #113, opened 2017-09-06
initial debianization

noxiouz:
- [x] Upstart/Systemd
- [x] Default configs?


debian/rules
```diff
@@ -0,0 +1,24 @@
...
+	dh $@ --with systemd
+
+override_dh_auto_build:
+	make -j build/insomnia
+
+override_dh_auto_test:
+	echo "skip tests"
+
+override_dh_auto_install:
+	install -d $(DEB_DESTDIR)etc/sonm/
```
antmat: Looks like you have only configs in your package

debian/rules
```diff
@@ -0,0 +1,24 @@
...
+
+
+%:
+	dh $@ --with systemd
+
+override_dh_auto_build:
+	make -j build/insomnia
+
+override_dh_auto_test:
+	echo "skip tests"
```
antmat: why?

debian/rules
```diff
@@ -0,0 +1,24 @@
...
+override_dh_auto_install:
+	install -d $(DEB_DESTDIR)etc/sonm/
+	install -m644 miner.yaml $(DEB_DESTDIR)etc/sonm/miner-default.yaml
+	install -m644 hub.yaml $(DEB_DESTDIR)etc/sonm/hub-default.yaml
+
+override_dh_systemd_start:
+	echo "Not running dh_systemd_start"
+
+override_dh_auto_clean:
+	rm -rf etc/
```
antmat: Are you sure?

noxiouz: Yep) The trick is not to forget / =)

debian/rules
noxiouz: requires lots of stuff to do. For example, generate mocks, which is tricky under debuild :)
At the same time we run tests in Travis before debuild

PR #127, opened 2017-09-23
feat: bid/ask hub protocol

3Hren:
Deeply WIP. But still can be merged to allow parallel work.


sshaman1101:
Thanks for your work! 
@iolebedev suggests to add one more param to `NetworkRequirements`:  Some kind of restrictions for network operations: 
1) Port for incoming connections is required
2) Outgoing connections are required
3) No network at all

Is it OK that {CPU, RAM, Disk,GPU}Requirements do not use as part of `BidItem` struct? :)

sshaman1101:
Maybe implement resource types as enums fields?
For example
```
message GPU {
    enum GpuType {
        ANY = 0;
    }
    GpuType gpuType = 13;
    uint64 gpuRam = 14;
}

message Order {
    enum CpuType {
        ANY = 0;
    }
    enum StorageType {
        OTHER = 0;
        SLOW_HDD = 1;
        FAST_HDD = 2;
        SSD = 3;
    }
    enum OrderType {
        BID = 0;
        ACK = 1;
    }
    enum NetworkMode {
        INCOMING = 0;
        OUTCOMING = 1;
        NONE = 2;
    }

    uint64 startTime = 1;
    uint64 endTime = 2;
    float buyerRating = 3;
    float supplierRating = 4;

    uint64 cpuQtty = 5;
    CpuType cpuType = 6;
    uint64 ramQtty = 7;

    uint64 storageQtty = 8;
    StorageType storageType = 9;

    uint64 networkQtty = 10;
    uint64 networkLat = 11;
    NetworkMode networkMode = 12;


    uint64 gpuQtty = 13;
    repeated GPU gpu = 14;

    float price = 15;
    OrderType orderType = 16;

    string id = 17;
    string supplierID = 18;
    string buyerID = 19;
}
```

Also, if we will keep this types as proto, not as golang constants, we can use this messages into another part of SONM (wallet, Marketplace).

3Hren:
I'm not sure about types, except order type. So making them as enums seems prematurely.

insonmnia/hardware/marshal.go
```diff
@@ -120,13 +121,19 @@ func GPUFromProto(g []*pb.GPUDevice) ([]*gpu.Device, error) {
 			cacheSize = 0
 		}
 
+		maxMemorySize, err := strconv.ParseUint(i.Ext["max_memory_size"], 10, 0)
+		if err != nil {
+			cacheSize = 0
```
sshaman1101: Why `cacheSize = 0` if you cannot parse `maxMemorySize`?

3Hren: Wow, typo.

3Hren: Fixed. Anyway I want to migrate to strictly-typed messages when we stabilize what we exactly want from devices.

3Hren:
Well, switched to strictly-typed GPU. Also some YAGNI.

insonmnia/hardware/marshal.go
```diff
@@ -101,34 +95,11 @@ func GPUIntoProto(g []*gpu.Device) []*pb.GPUDevice {
 	return result
 }
 
-func GPUFromProto(g []*pb.GPUDevice) ([]*gpu.Device, error) {
-	result := make([]*gpu.Device, 0)
+func GPUFromProto(g []*pb.GPUDevice) ([]gpu.Device, error) {
+	result := []gpu.Device{}
```
noxiouz: capacity is known before (len(g))?

PR #129, opened 2017-09-28
Update Bid-Ask protocol to latest version

sshaman1101:


proto/bid.proto
```diff
@@ -9,39 +9,61 @@ enum OrderType {
...
+}
+
+message Slot {
     // Virtual computer start of life (hour grained).
     Timestamp startTime = 1;
     // Virtual computer end of life (hour grained).
     Timestamp endTime = 2;
     // Buyer’s rating. Got from Buyer’s profile for BID orders rating_supplier.
-    float ratingBuyer = 3;
+    float buyerRating = 3;
```
3Hren: I'd rather made this non-float to avoid FP loss.

PR #130, opened 2017-10-02
Aux/marketplace

sshaman1101:
This PR adds a simple centralized marketplace implementation.


insonmnia/marketplace/server.go
```diff
@@ -0,0 +1,263 @@
...
+		return errSlotIsNil
+	}
+
+	if slot.Resources == nil {
+		return errResourcesIsNil
+	}
+	return nil
+}
+
+func (in *inMemOrderStorage) compareTime(slot *pb.Slot, order *pb.Order) bool {
```
3Hren: Order contains a slot, which means that you're comparing a slot with another slot. Then maybe better to represent it explicitly?

`func (s *Slot) compareTime(other *Slot) bool` and so on

3Hren: Of course you need a wrapper type: `type Slot pb.Slot`.

3Hren:
I need the same code for comparing slots, so I'm going to just wait until you complete :)

insonmnia/marketplace/server.go
```diff
@@ -0,0 +1,270 @@
...
+	errOrderIsNil        = errors.New("Order cannot be nil")
+	errSlotIsNil         = errors.New("Order slot cannot be nil")
+	errResourcesIsNil    = errors.New("Slot resources cannot be nil")
+	errStartTimeAfterEnd = errors.New("Start time is after end time")
+	errStartTimeRequired = errors.New("Start time is required")
+	errEndTimeRequired   = errors.New("End time is required")
+)
+
+type OrderStorage interface {
+	GetOrders(slot *pb.Slot) ([]*pb.Order, error)
```
3Hren: Great idea: make `Order` wrapper, which is constructed from `pb.Order` and performs all required checks (like required fields). Otherwise all other implementation will suffer from doing the same things.

Protobuf sucks, yeah.

Make in a separate PR.

3Hren:
Also conflicts. Rebase and merge.

3Hren:
Emm, why did you add locator code here? It is out of the scope and is discussed in a separate PR.

sshaman1101:
Oh, merged with my local changes and accidentally pushed.
Commits from the locator branch were removed!

PR #131, opened 2017-10-06
aux/update-deb

sshaman1101:
This PR updates a Debian packages version; 
Also, adds deb package with the Marketplace.


debian/changelog
```diff
@@ -1,5 +1,5 @@
-sonm-core (0.1.1) unstable; urgency=low
```
3Hren: Changelog is for appending, not overwriting :)

PR #140, opened 2017-10-18
Feature/local node impl

sshaman1101:
This PR adds some local node methods.
Currently, we can implement Hub management operations and Marketplace operations.

TODO:

**Hub:**
- [x] Status

Workers:
- [x] WorkerList
- [x] WorkerStatus
- [x] GetWorkerProperties
- [x] SetWorkerProperties

ACLs:
- [x] Add stubs on Hub
- [x] GetRegistredWorkers
- [x] RegisterWorker
- [x] UnregisterWorker

Orders:
- [x] GetPlanByID
- [x] CreateAskPlan
- [x] DeleteAskPlan

Tasks:
- [x] TaskList
- [x] TaskStatus

**Market:**
- [x] GetOrders
- [x] GetOrderByID
- [x] CreateOrder (Bid only)
- [x] CancelOrder (Bid only)


sshaman1101:
Implemented LocalNode methods for a ready to use parts of the SONM;
I see no reason to add not implemented commands to the CLI now (tasks and deals).

After this PR will be merged I would tune some inner logic into the Marketplace service.

@noxiouz @3Hren @antmat PTAL

Makefile
```diff
@@ -82,6 +86,10 @@ install/cli: build/cli
 	@echo "+ $@"
 	cp ${CLI} ${INSTALLDIR}
 
+install/node: build/node
+	@echo "+ $@"
+	cp ${CLI} ${LOCAL_NODE}
```
antmat: copy paste?

cmd/cli/commands/miner.go
```diff
@@ -122,7 +115,7 @@ var minerStatusCmd = &cobra.Command{
 	PreRunE: checkHubAddressIsSet,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if len(args) < 1 {
-			return errMinerAddressRequired
+			return errWorkerAddressRequired
```
antmat: if you already started renaming - how about something like 'errWorkerIdRequired' ?

hub.yaml
```diff
@@ -1,13 +1,13 @@
 # An endpoint for miner connections.
 # allowed format is "ip:port" to bind Hub to given IP address
 # or ":port" to bind Hub to all available IPs
-endpoint: ":10002"
+endpoint: "172.16.1.118:10002"
```
antmat: I don't think it's a good idea)

hub.yaml
```diff
@@ -20,7 +20,7 @@ monitoring:
   # An endpoint for client connections using cli
   # allowed format is "ip:port" to bind Hub to given IP address
   # or ":port" to bind Hub to all available IPs
-  endpoint: ":10001"
+  endpoint: "172.16.1.118:10001"
```
antmat: here too

insonmnia/node/common.go
```diff
@@ -0,0 +1,18 @@
...
+
+import "google.golang.org/grpc"
+
+// credentials describe any type of auth data.
+// Empty for now, but must be implemented in nearest future
+type credentials interface{}
+
+// initGrpcClient creates new client connection on addr and wraps it with credentials
+func initGrpcClient(addr string, _ credentials) (*grpc.ClientConn, error) {
+	cc, err := grpc.Dial(addr, grpc.WithInsecure(),
```
antmat: Another copy of this code. How about making a library method for connecting with sonm specific parameters?

insonmnia/node/hub.go
```diff
@@ -1,52 +1,156 @@
...
 import (
+	log "github.com/noxiouz/zapctx/ctxlog"
 	pb "github.com/sonm-io/core/proto"
 	"golang.org/x/net/context"
 )
 
-type hubAPI struct{}
+type hubAPI struct {
+	conf Config
+	cc   pb.HubClient
```
antmat: why 'cc'?

insonmnia/node/hub.go
```diff
@@ -1,52 +1,156 @@
...
+	ctx  context.Context
+}
+
+func (h *hubAPI) Status(ctx context.Context, req *pb.Empty) (*pb.HubStatusReply, error) {
+	log.G(h.ctx).Debug("handling Status request")
```
antmat: Should not it be 'Info'?

insonmnia/node/hub.go
```diff
@@ -1,52 +1,156 @@
...
+
+	workers, err := h.getWorkersIDs(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	slots := []*pb.Slot{}
+	for _, wrkID := range workers {
+		req := &pb.ID{Id: wrkID}
+		workerSlots, err := h.cc.GetSlots(ctx, req)
```
antmat: looks like this can be done on hub via one call(which would be more efficient)

insonmnia/node/hub.go
```diff
@@ -1,52 +1,156 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	reply := &pb.TaskListReply{
+		Info: map[string]*pb.TaskListReply_TaskInfo{},
+	}
+
+	for _, wrkID := range workers {
+		taskStatuses, err := h.cc.MinerStatus(ctx, &pb.ID{Id: wrkID})
```
antmat: here too - it looks like it would be better to add aggregating method to hub

node.yaml
```diff
@@ -0,0 +1,21 @@
...
+# Marketplace service settings
+market:
+  # marketplace gRPC endpoint
+  endpoint: 127.0.0.1:9095
+
+# Hub management settings
+hub:
+  # endpoint is Hub's gRPC endpoint
+  # for client connections
+  endpoint: 172.16.1.118:10001
```
antmat: comment this and replace with some example like 1.2.3.4

cmd/cli/commands/common.go
```diff
@@ -70,6 +83,13 @@ func checkHubAddressIsSet(cmd *cobra.Command, _ []string) error {
 	return nil
 }
 
+func checkNodeAddressIsSet(cmd *cobra.Command, _ []string) error {
```
antmat: Why do you need this when node address has default value?

cmd/cli/commands/interactor.go
```diff
@@ -149,3 +149,224 @@ func NewGrpcInteractor(addr string, to time.Duration) (CliInteractor, error) {
...
+	_, err := it.market.CancelOrder(ctx, &pb.Order{Id: id})
+	return err
+}
+
+func (it *marketInteractor) ctx() (context.Context, context.CancelFunc) {
+	return context.WithTimeout(context.Background(), it.timeout)
+}
+
+func NewMarketInteractor(addr string, timeout time.Duration) (NodeMarketInteractor, error) {
+	cc, err := grpc.Dial(addr, grpc.WithInsecure(),
```
antmat: again same code (and in many other places)

cmd/cli/commands/node_order.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func loadOrderFile(path string) (*structs.Order, error) {
+	buf, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	cfg := task_config.OrderConfig{}
+	err = yaml.Unmarshal(buf, &cfg)
```
antmat: This code looks very similar to previous function (damn go with no generics!) 
Can we do something with it? something like loadYamlFileToStruct(path string, to interface{}) ?

cmd/cli/commands/node_tasks.go
```diff
@@ -0,0 +1,88 @@
...
+				cmd.Printf("Worker \"%s\" has no tasks\r\n", worker)
+				continue
+			}
+
+			cmd.Printf("Worker \"%s\":\r\n", worker)
+			i := 1
+			for ID, status := range tasks.GetTasks() {
+				up := time.Duration(status.GetUptime())
+				cmd.Printf("  %d) %s \r\n     %s  %s (up: %v)\r\n",
+					i, ID, status.Status.String(), status.ImageName, up.String())
```
antmat: i++ in param?

cmd/cli/commands/node_worker.go
```diff
@@ -0,0 +1,151 @@
...
+
+		_, err = hub.SetWorkerProperties(req)
+		if err != nil {
+			showError(cmd, "Cannot get workers status", err)
+			os.Exit(1)
+		}
+	},
+}
+
+func parsePropsFile(path string) (map[string]string, error) {
```
antmat: again code duplication

cmd/cli/task_config/order.go
```diff
@@ -0,0 +1,50 @@
...
+		ByuerID:    c.BuyerID,
+		SupplierID: c.SupplierID,
+		Price:      c.Price,
+		OrderType:  orderType,
+		Slot:       slot.Unwrap(),
+	})
+}
+
+func ParseOrderType(ty string) (sonm.OrderType, error) {
+	switch ty {
```
antmat: I think this should be in structs package (and it can use generated protobuf map)

cmd/node/main.go
```diff
@@ -0,0 +1,49 @@
...
+
+	n, err := node.New(ctx, cfg)
+	if err != nil {
+		fmt.Printf("Err: cannot build Node instance: %s\r\n", err)
+		os.Exit(1)
+	}
+
+	fmt.Printf("Starting Local Node at %s...\r\n", cfg.ListenAddress())
+	if err := n.Serve(); err != nil {
+		fmt.Printf("Cannot start Local Node: %s\r\n", err.Error())
```
antmat: If some error happens during n.Serve() message will be a little misleading.

cmd/cli/commands/node_tasks.go
sshaman1101: in param for what?

antmat: yep, you are right, go doesn't allow this

cmd/node/main.go
sshaman1101: Now `.Serve()` method is just for serving :)

3Hren:
It's too hard to review such huge PR's. Either we review all and the entire process takes months, or merge it with hand-on-heart with further quick fix.

cmd/node/main.go
antmat: I mean node is working for example for 1 day. After that it returns err and we see "Cannot start Local Node"

PR #144, opened 2017-10-21
Fix/slot comparison

sshaman1101:
**This PR is based on #140, please merge it first.**

This PR changing a GPU devices representation from "N devices is required" to "no or one or many devices is required".

Also enabled gRPC compression for marketplace service;


3Hren:
Rebase to eliminate commits associated with previous PR.

Issue #142, opened 2017-10-19
Hub still hangs forever on SIGINT


3Hren:
Probably, in the consul.

antmat:
Not really. It seems we even do not call cancellation function for context (and it hangs forever in startLocatorAnnouncer function)

PR #149, opened 2017-10-24
fix: hanging on the stop

antmat:
fixes https://github.com/sonm-io/core/issues/142


insonmnia/hub/server.go
```diff
@@ -660,6 +661,7 @@ func (h *Hub) RemoveSlot(ctx context.Context, request *pb.RemoveSlotRequest) (*p
 
 // New returns new Hub
 func New(ctx context.Context, cfg *HubConfig, version string) (*Hub, error) {
+	ctx, cancel := context.WithCancel(ctx)
```
noxiouz: Move it as much close as possible to &Hub to avoid cancel leak in err != nil :(

PR #157, opened 2017-11-01
feature: background order handler for the Node

sshaman1101:
### Node
Added: 
- order handler for executing orders from the Marketplace (search -> propose -> deal)

### CLI:
Added:
- Method for inspecting processing orders from the Node


insonmnia/marketplace/server.go
```diff
@@ -199,6 +196,10 @@ func (m *Marketplace) CancelOrder(_ context.Context, req *pb.Order) (*pb.Empty,
 	return &pb.Empty{}, nil
 }
 
+func (m *Marketplace) GetProcessing(ctx context.Context, req *pb.Empty) (*pb.GetProcessingReply, error) {
+	return nil, nil
```
Sokel: `notImplemented` maybe?

sshaman1101: It exists just for matching the Marketplace interface. The Market service itself unable to know anything about processing orders, this method is for the Node only.

sshaman1101: But you're right, I must place some note about this into the code.

insonmnia/node/market.go
```diff
@@ -1,18 +1,219 @@
...
+
+const (
+	statusNew uint8 = iota
+	statusSearching
+	statusProposing
+	statusDealing
+	statusDone
+	statusFailed
+
+	orderPoolPeriod = 5 * time.Second
```
3Hren: `s/Pool/Poll` maybe?

PR #160, opened 2017-11-02
opencl: support OpenCL API

noxiouz:
There are 3 modes now: *nil*, *nvidiadocker*, *embedded*:

+ *nil*: just a placeholder for unsupported platforms or if the feature disabled
+ *nvidiadocker*: old-style way to provide GPU support with an extra fix to enable OpenCL for Nvidia.  Works via NvidiaDocker plugin
+ *embedded*: supports both nvidia (GPU/OpenCL) and ATI OpenCL. Parts of NvidiaDocker are embedded into a miner with some improvements (i.e volume API is rewritten). Miner acts as docker volume driver to bind the libraries inside a container. 

**clinfo**  util was used as testing utility

**NOTE**: a container should have LD_LIBRARY_PATH pointing to /usr/local/nvidia/lib{64} and PATH /usr/local/nvidia/bin. Base images for nvidia-docker contain such env fixes.
Although we can set LD_LIBRARY_PATH from outside the container, it's impossible to append into PATH variable

### DIFF
--- a/cmd/miner/main.go
+++ b/cmd/miner/main.go
@@ -14,9 +14,10 @@ import (
 	"github.com/sonm-io/core/insonmnia/logging"
 	"github.com/sonm-io/core/insonmnia/miner"
 
+	"io/ioutil"
+
 	log "github.com/noxiouz/zapctx/ctxlog"
 	"github.com/pborman/uuid"
-	"io/ioutil"
 )
 
 var (
--- a/insonmnia/miner/config.go
+++ b/insonmnia/miner/config.go
@@ -45,7 +45,7 @@ type config struct {
 	HubConfig      *HubConfig      `required:"false" yaml:"hub"`
 	FirewallConfig *FirewallConfig `required:"false" yaml:"firewall"`
 	Eth            *EthConfig      `yaml:"ethereum"`
-	GPUConfig      *GPUConfig      `required:"false" yaml:"GPUConfig"`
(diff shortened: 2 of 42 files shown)


insonmnia/gpuisolation/graphic_driver_volume.go
```diff
@@ -0,0 +1,164 @@
...
+func NewPlugin(volumes nvidia.VolumeMap) volume.Driver {
+	return &volumePlugin{Volumes: volumes}
+}
+
+type volumePlugin struct {
+	Volumes nvidia.VolumeMap
+}
+
+func (p *volumePlugin) Create(req *volume.CreateRequest) error {
+	log.Printf("Received create request for volume '%s'", req.Name)
```
3Hren: Why'd not pass our famous `log.G(ctx)...` here and elsweyr?

insonmnia/gpuisolation/nvidia_init.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if _, err := os.Stat(openCLVendorDir); err == nil {
+		ovs.OpenCLVendorDir = openCLVendorDir
+	}
+
+	// NOTE: Add libraries we need to discover
+	nvolumes := nvidia.Volumes[0]
+	components := nvolumes.Components
+	components["libraries"] = append(components["libraries"],
+		"libdrm.so",
```
3Hren: Seems constant and platform-dependent.

PR #154, opened 2017-10-31
secure: grpc on https

noxiouz:

PR #161, opened 2017-11-05
feat: worker ACL

3Hren:
This PR implements worker ACL registration/deregistration.

Added:
- ACL storage on the Hub.
- New CLI commands: `acl register CREDENTIALS` and `acl deregister CREDENTIALS`.

Changed:
- Implemented above methods on the Hub.

Things to be done:
- Waiting for #154, since from there we can extract worker credentials from the certificate and to perform validation while establishing a worker connection.

### DIFF
--- a/insonmnia/hub/server.go
+++ b/insonmnia/hub/server.go
@@ -2,6 +2,7 @@ package hub
 
 import (
 	"crypto/ecdsa"
+	"crypto/x509"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -839,6 +840,13 @@ func New(ctx context.Context, cfg *HubConfig, version string) (*Hub, error) {
 		return nil, err
 	}
 
+	acl := NewACLStorage()
+	privateKey, err := x509.MarshalECPrivateKey(ethKey)
+	if err != nil {
+		return nil, err
+	}
+	acl.Insert(hex.EncodeToString(privateKey))
+
 	h := &Hub{
 		ctx:          ctx,
 		cancel:       cancel,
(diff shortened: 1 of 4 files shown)


insonmnia/hub/server.go
```diff
@@ -768,16 +775,18 @@ func (h *Hub) GetRegisteredWorkers(ctx context.Context, empty *pb.Empty) (*pb.Ge
...
 }
 
 // DeregisterWorkers deny Worker with given ID to connect to the Hub
 func (h *Hub) DeregisterWorker(ctx context.Context, request *pb.ID) (*pb.Empty, error) {
-	// todo: implement me
 	log.G(h.ctx).Info("handling DeregisterWorker request", zap.String("id", request.GetId()))
-	return nil, ErrUnimplemented
+
+	h.acl.Remove(request.Id)
+	return &pb.Empty{}, nil
```
sshaman1101: Maybe it will be better to check if the key is present in the storage, then delete it. 
If there is no key, then fire an error.

insonmnia/hub/server.go
```diff
@@ -831,6 +840,13 @@ func New(ctx context.Context, cfg *HubConfig, version string) (*Hub, error) {
 		return nil, err
 	}
 
+	acl := NewACLStorage()
+	privateKey, err := x509.MarshalECPrivateKey(ethKey)
+	if err != nil {
+		return nil, err
+	}
+	acl.Insert(hex.EncodeToString(privateKey))
```
sshaman1101: For what do you add Hub's key in itself's Storage? 
There is ACLs for Workers, not for Hubs, isn't it?

3Hren: As we discussed this is default value to enable key sharing between hub and workers.

PR #162, opened 2017-11-06
Feat: kv absatraction

antmat:
this PR adds docker/libkv and docker/leadership abstraction to hide consul implementation details.
@3Hren @noxiouz @sshaman1101 PTAL

Upd:
Features:
  * Use libkv as abstraction on top of consul:
    * To provide more simple interface for leader election
    * Add ability to use zookeeper and etcd as well
    * Add ability to use boltdb (for functional tests f.e.)
  * Extract cluster entity outside Hub:
    * Split discovery and leader election to separate entities
    * Use all possible endpoints for discovery instead of one
    * Add generalized mechanism to register synchronizable entities, watch for their changes, and emit clusterEvents on their change
    * Incapsulate cross hub grpc connectivity inside cluster
    * Use full mesh interconnectivity instead of only leader client
    * Watcher for dropping connection to stale nodes
  * Adapt hub server to new cluster entity

TODO:
  * [x] React on err from cluster in eventChannel


insonmnia/hub/server.go
```diff
@@ -873,47 +868,39 @@ func New(ctx context.Context, cfg *HubConfig, version string) (*Hub, error) {
...
+			h.leaderClient = nil
+			h.leaderClientLock.Unlock()
+
+			h.onNewHub(leader)
+			conn, err := grpc.Dial(leader, grpc.WithInsecure(),
+				grpc.WithCompressor(grpc.NewGZIPCompressor()),
+				grpc.WithDecompressor(grpc.NewGZIPDecompressor()))
+			if err != nil {
+				log.G(h.ctx).Warn("could not connect to hub", zap.String("endpoint", leader), zap.Error(err))
+				time.Sleep(time.Duration(100 * 1000000))
```
noxiouz: 100 * time.Second ?

insonmnia/hub/server.go
```diff
@@ -873,47 +868,39 @@ func New(ctx context.Context, cfg *HubConfig, version string) (*Hub, error) {
...
+		case <-h.ctx.Done():
+			return nil
+		case err := <-errCh:
+			log.G(h.ctx).Error("leader watch failed", zap.Error(err))
+			h.Close()
+			return err
+		case leader := <-leaderCh:
+			log.G(h.ctx).Info("leader watch: fetched leader", zap.Any("leader", leader))
+			h.leaderClientLock.Lock()
+			h.leaderClient = nil
```
noxiouz: is it possible to have grpc connection leak?

3Hren:
It is not what we meant.

This PR does not gives an abstraction, it's just another decomposition a raw KV storage into some kind of new storage. It does not gives us a way to add new synchronizable items across Hub cluster nor to fetch them.

For example there is an interface of such cluster synchronization mechanism.

```go
// ClusterEvent describes an event that can produce the cluster.
//
// Possible types are:
// - `map[string]DeviceProperties` when received device properties updates.
// - `struct{}` when switched the state.
// - `error` when a connection to the Consul is broken.
type ClusterEvent interface{}

type Cluster interface {
	// IsLeader returns true if this cluster is a leader, i.e. we rule the
	// synchronization process.
	IsLeader() bool
	
	// All these operations should fail if this node is not a leader.
	
	// SynchronizeDevices synchronizes device properties with followers.
	SynchronizeDevices(properties map[string]DeviceProperties) error
	// ... and so on for other stuff we need to synchronize.
	SynchronizeTasks(...)
	SynchronizeSlots(...)
	SynchronizeACL(...)
}

// Returns a cluster writer interface if this node is a master, event channel
// otherwise.
// Should be recalled when a cluster's master/slave state changes.
// The channel is closed when the specified context is canceled.
func NewCluster(ctx context.Context, consul Consul) (Cluster, chan ClusterEvent, error)
```

insonmnia/hub/cluster.go
```diff
@@ -0,0 +1,396 @@
...
+			} else {
+				for _, member := range members {
+					err := c.registerMember(member)
+					if err != nil {
+						log.G(c.ctx).Warn("trash data in cluster members folder: ", zap.Any("kvPair", member))
+					}
+				}
+			}
+		case <-c.ctx.Done():
+			stopCh <- struct{}{}
```
noxiouz: Safer just close this chan to notify

insonmnia/hub/cluster.go
```diff
@@ -0,0 +1,396 @@
...
+func (c *cluster) watchEvents() {
+	watchStopChannel := make(chan struct{})
+	ch, err := c.store.WatchTree(synchronizableEntitiesPrefix, watchStopChannel)
+	if err != nil {
+		c.close(err)
+		return
+	}
+	for {
+		select {
+		case <-c.ctx.Done():
```
noxiouz: return ?

insonmnia/hub/cluster.go
```diff
@@ -0,0 +1,396 @@
...
+			c.close(err)
+		case c.leaderId = <-leaderCh:
+			c.emitLeadershipEvent()
+		}
+	}
+}
+
+func (c *cluster) announce() {
+	endpointsData, _ := json.Marshal(c.endpoints)
+	ticker := time.NewTicker(time.Second * 1)
```
noxiouz: defer tiecker.Stop()

insonmnia/hub/cluster.go
```diff
@@ -0,0 +1,396 @@
...
+				if err != nil {
+					c.eventChannel <- value.Interface()
+				}
+			}
+		}
+	}
+}
+
+func (c *cluster) nameByEntity(entity interface{}) (string, error) {
+	c.registeredEntitiesMu.RLock()
```
noxiouz: defer RUnlock

insonmnia/hub/cluster.go
```diff
@@ -0,0 +1,396 @@
...
+	t := reflect.TypeOf(entity)
+	name, ok := c.entityNames[t]
+	if !ok {
+		return "", errors.New("entity " + t.String() + " is not registered")
+	}
+	return name, nil
+}
+
+func (c *cluster) typeByName(name string) (reflect.Type, error) {
+	c.registeredEntitiesMu.RLock()
```
noxiouz: defer RUnlock

insonmnia/hub/cluster.go
```diff
@@ -0,0 +1,396 @@
...
+		}
+	}
+
+}
+
+func (c *cluster) hubWatch() {
+	stopCh := make(chan struct{})
+	listener, err := c.store.WatchTree(listKey, stopCh)
+	if err != nil {
+		c.close(err)
```
noxiouz: return?

insonmnia/hub/cluster.go
```diff
@@ -0,0 +1,396 @@
...
+		return err
+	}
+	for _, ep := range endpoints {
+		conn, err := util.MakeGrpcClient(ep, nil)
+		if err != nil {
+			log.G(c.ctx).Warn("could not connect to hub", zap.String("endpoint", ep), zap.Error(err))
+			continue
+		} else {
+			c.leaderLock.Lock()
+			c.clients[id] = pb.NewHubClient(conn)
```
noxiouz: how are going to close conn?

insonmnia/hub/cluster.go
```diff
@@ -0,0 +1,396 @@
...
+
+	config := store.Config{}
+	config.Bucket = cfg.StoreBucket
+	return libkv.NewStore(backend, endpoints, &config)
+}
+
+func (c *cluster) close(err error) {
+	log.G(c.ctx).Error("cluster failure", zap.Error(err))
+	c.eventChannel <- err
+	close(c.eventChannel)
```
noxiouz: Not sure, but it seems u may close this channel more than once

insonmnia/hub/cluster.go
```diff
@@ -0,0 +1,496 @@
+package hub
+
+import (
+	"context"
```
3Hren: This import order is blatantly not sorted.

insonmnia/hub/cluster.go
```diff
@@ -0,0 +1,496 @@
...
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"net"
+	"reflect"
+	"strings"
+	"sync"
+	"time"
+)
+
+const leaderKey = "sonm/hub/leader"
```
3Hren: Configurable?

insonmnia/hub/cluster.go
antmat: Huh?

insonmnia/hub/cluster.go
antmat: Not sure if we could ever need this

insonmnia/hub/cluster.go
3Hren: goimport

insonmnia/hub/cluster.go
antmat: Anyway, let's do it in separate PR

insonmnia/hub/cluster.go
```diff
@@ -0,0 +1,507 @@
...
+			c.leaderLock.Unlock()
+			c.emitLeadershipEvent()
+		}
+	}
+}
+
+func (c *cluster) announce() {
+	log.G(c.ctx).Info("starting announce goroutine", zap.Any("endpoints", c.endpoints), zap.String("ID", c.id))
+	endpointsData, _ := json.Marshal(c.endpoints)
+	ticker := time.NewTicker(time.Second * 10)
```
noxiouz: defer ticker.Stop()

insonmnia/hub/cluster.go
```diff
@@ -0,0 +1,507 @@
...
+			cli.conn.Close()
+			delete(c.clients, id)
+		}
+	}
+	return nil
+}
+
+func (c *cluster) hubGC() {
+	log.G(c.ctx).Info("starting hub GC goroutine")
+	t := time.NewTicker(time.Second * 60)
```
noxiouz: defer t.Stop()

insonmnia/hub/server.go
```diff
@@ -1118,10 +967,40 @@ func (h *Hub) Serve() error {
...
+			h.processClusterEvent(event)
+		case <-h.ctx.Done():
+			return
+		}
+	}
+}
+
+func (h *Hub) processClusterEvent(value interface{}) {
+	log.G(h.ctx).Info("received cluster event", zap.Any("event", value))
+	switch value.(type) {
```
noxiouz: switch value := value.(type)

in that case, you can omit type case in case *map[string]*TaskInfo

antmat:
Merging as @noxiouz approved

PR #166, opened 2017-11-08
Feature: key loader

sshaman1101:
This PR describes an interface for opening Etherum keys.

Added:
- `KeyOpener` interface and base implementation
- `PassPhraser` interface and implementations for interactive prompting (using [gopass]("github.com/howeyc/gopass"))

TBD:
- after review/merge, we will add keys loading for the cli and the local node.


accounts/identity.go
```diff
@@ -157,12 +149,12 @@ func parseKeystoreUrl(path string) (string, error) {
 	return u.Path, nil
 }
 
-// return default keystore directory stored in in `.sonm` directory
-// if any error occurred .sonm directory will be in working dir
+// GetDefaultKeystoreDir return default keystore directory places in "~/.sonm/keystore".
+// If any error occurred ".sonm" directory will be in current working dir
 func GetDefaultKeystoreDir() string {
-	rootDir, err := util.GetUserHomeDir()
+	homeDir, err := util.GetUserHomeDir()
```
3Hren: I'd rather fail completely in this case instead of magically creating `.sonm` directory whenever an user happened to run this code.

accounts/keys.go
```diff
@@ -0,0 +1,168 @@
...
+	"crypto/ecdsa"
+	"errors"
+	"io"
+	"os"
+
+	"github.com/howeyc/gopass"
+)
+
+var (
+	errNoKeystoreDir = errors.New("Keystore directory does not exists")
```
3Hren: Usually error descriptions are written starting with lowercase to be able to chain them. For example such description with error formatted as `"failed to do magic: %s"` will emit "failed to do magic: Keystore directory does not exists".

accounts/keys.go
```diff
@@ -0,0 +1,168 @@
...
+	GetPassPhraser() PassPhraser
+	// OpenKeystore opens key storage.
+	OpenKeystore() (bool, error)
+	// GetKey returns private key from opened storage
+	GetKey() (*ecdsa.PrivateKey, error)
+}
+
+// nullPassPhraser implements PassPhraser and always return the same pass phrase.
+// NOTE: Use it for testing purposes only!
+type nullPassPhraser struct{}
```
3Hren: Mock instead.

accounts/keys.go
```diff
@@ -0,0 +1,168 @@
...
+
+	idt = NewIdentity(o.keyDirPath)
+	err = idt.Open(passPhrase)
+	if err == nil {
+		return false, nil
+	}
+
+	if err == ErrWalletIsEmpty {
+		_, err = o.createNewKey(idt)
+		return err == nil, err
```
3Hren: It took me about 5 seconds to figure out what's going on. This is too complicated for typical go developer.

accounts/keys.go
```diff
@@ -0,0 +1,168 @@
...
+	GetPassPhrase() (string, error)
+}
+
+// KeyOpener is interface for loading Eth keys
+type KeyOpener interface {
+	// GetPassPhraser return PassPhraser interface
+	// that provides pass phrase for loaded keys
+	GetPassPhraser() PassPhraser
+	// OpenKeystore opens key storage.
+	OpenKeystore() (bool, error)
```
3Hren: Why is there a bool result? "We couldn't open a keystore, but whatever, we don't know how it happened"?

accounts/keys.go
```diff
@@ -0,0 +1,168 @@
...
+
+	return false, nil
+}
+
+func (o *defaultKeyOpener) GetKey() (*ecdsa.PrivateKey, error) {
+	if o.idt == nil {
+		return nil, ErrWalletNotOpen
+	}
+
+	key, err := o.idt.GetPrivateKey()
```
3Hren: `return o.idt.GetPrivateKey()`?

accounts/keys.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = idt.Open(passPhrease)
+	if err != nil {
+		return nil, err
+	}
+
+	key, err := idt.GetPrivateKey()
```
3Hren: `return idt.GetPrivateKey()`?

accounts/keys.go
sshaman1101: A bool flag shows that keystore was created (false if just opened). Need to documenting that.

accounts/identity.go
sshaman1101: Looks like this function used only for testing and into the examples. 

@Sokel is it OK to move this code somewhere into the `blockchain/example/`?

PR #170, opened 2017-11-09
Feature/cli load keys

sshaman1101:
This PR adds the `login` method for CLI. The method allows creating or opening an Etherum keystore.

TBD:
- [x] extract key loading logic implemented in loginCmd command;
- [x] Add a "middleware" (using _cobra.Command.PreRunE_) to load the storage on each action that must be signed;

### DIFF
--- a/cmd/cli/commands/common.go
+++ b/cmd/cli/commands/common.go
@@ -59,6 +59,8 @@ func init() {
 
 	nodeRootCmd.AddCommand(nodeHubRootCmd, nodeMarketRootCmd)
 	rootCmd.AddCommand(hubRootCmd, minerRootCmd, tasksRootCmd, versionCmd, nodeRootCmd)
+
+	rootCmd.AddCommand(loginCmd)
 }
 
 var nodeRootCmd = &cobra.Command{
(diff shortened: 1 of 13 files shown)


sshaman1101:
@noxiouz @3Hren PTAL

cmd/cli/commands/common.go
```diff
@@ -151,3 +156,40 @@ func isSimpleFormat() bool {
 
 	return true
 }
+
+// printer interface describe anything that can print
+// something somehow on a something.
```
3Hren: Such abstraction
So java
Much whatever
Wow.

util/util.go
```diff
@@ -14,6 +14,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/crypto"
 	"gopkg.in/yaml.v2"
+	"os"
```
3Hren: Sort.

util/util.go
```diff
@@ -116,3 +117,11 @@ func LoadYamlFile(from string, to interface{}) error {
 
 	return nil
 }
+
+// DirectoryExists returns true if given directory exists
```
3Hren: The given.

PR #180, opened 2017-11-14
feat: associate deal with the tasks

3Hren:
This PR adapts existing task API to work with deals. An intermediate
cgroup layer also introduced, making possible to limit resources for each
deal entirely by chaining cgroups in a three level structure: parent
group for entire miner <- group for a deal <- group for a container.

## Added:
- Resource consumption over an allocated slot can be asked from a miner.
- Parent cgroup name can now be configured from a config.
- Nested cgroup manager to be able to keep and remember about named cgroups.
  These cgroups are owned by the parent group, which makes possible to set
  limits for a deal. Tasks that are bound to such deals will have their
  own cgroups backed with that intermediate group, which results in three
  layer control group.
- Add tests for parsing docker port binding.

## Changed:
- Parent cgroup can be disabled even on linux. This makes possible to
  run cgroup-related tests without sudo.
- Deal prototypes moved in its own file.
- Starting a task now requires deal specification.
- Check contract using ethereum. Currently ETH client is stubbed.
- Task info now holds the wrapped request.
- The proper miner lookup for starting a task is now performed by
  matching with an order id.
- Route registration is now in the miner context.
- Extracted parsing Docker port binding into a separate module.
- Mapping docker port from string to integer is now being done at the
  miner's side.
- Eliminated inner struct for miner start reply.
- Hide formatting env over its own wrapper.
- Miner start task request now has an order id to be able to associate
  resource allocation and limitations with that order.
- CGroup deleter now becomed just a tiny CGroup wrapper with an ability
  to attach nested control groups.
- Environment formatting is now done in the container description.

## Removed:
- Obsolete scheduling specific code were removed.


insonmnia/miner/overseer.go
```diff
@@ -34,13 +35,22 @@ type Description struct {
...
+	Env           map[string]string
 	TaskId        string
 	CommitOnStop  bool
 
 	GPURequired bool
 }
 
+func (d *Description) FormatEnv() []string {
+	vars := make([]string, 0, len(map[string]string(d.Env)))
+	for k, v := range map[string]string(d.Env) {
```
sshaman1101: Is there any reason to wrap `d.Env` (type `map[string]string`) with another `map[string]string` ? 
Also, it works OK if `d.Env` is nil map.

3Hren:
This is quite large change, so at least 2/3 reviewers are required.

cmd/cli/commands/tasks.go
```diff
@@ -285,17 +285,7 @@ func taskStartCmdRunner(cmd *cobra.Command, taskConfig task_config.TaskConfig, i
 		cmd.Printf("Starting \"%s\" ...\r\n", taskConfig.GetImageName())
 	}
 
-	requirements := &pb.TaskRequirements{
```
3Hren: For now we've discussed that there can be only one task per-deal.

insonmnia/structs/resources.go
```diff
@@ -3,9 +3,17 @@ package structs
 import (
 	"reflect"
 
+	"github.com/docker/docker/api/types/container"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
```
sshaman1101: Just `"errors"`?

insonmnia/hub/server.go
```diff
@@ -165,105 +163,18 @@ func (h *Hub) Info(ctx context.Context, request *pb.ID) (*pb.InfoReply, error) {
...
-	for _, filter := range h.filters {
-		ok, err := filter(miner, req)
-		if err != nil {
-			return false, err
-		}
-
-		if !ok {
-			return false, nil
-		}
+func (h *Hub) onRequest(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
```
antmat: Strange diff. Did you change anything in onRequest?

insonmnia/structs/task.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/sonm-io/core/proto"
+)
+
+var (
+	errDealRequired  = errors.New("deal is required")
+	errBidIdRequired = errors.New("bid id must be non-empty")
+)
+
+type StartTaskRequest struct {
+	*sonm.HubStartTaskRequest
```
antmat: we use pb as sonm alias everywhere. I Iike `sonm.` but I vote for consistency

insonmnia/miner/overseer.go
3Hren: No reason. One day it was a real wrapper, not - it's not.

3Hren:
> Looks great, now try to build and test it. 

?

insonmnia/structs/task.go
3Hren: I like `sonm` too.

But it's module local change, it does not affect aliases in other modules. In case of rename-everything a strongly disrespect such changes, but for migration state it's ok.

insonmnia/hub/server.go
3Hren: Nope.

3Hren:
> ... maybe I will take another look som time later

Ok, I'll wait a couple of months.

antmat:
Deal!

3Hren:
Timeout.

PR #182, opened 2017-11-17
fix: sync entities in cluster

antmat:
This PR enables cluster synchronization for acls, slots and deviceProperties.
Also in the sake of performance (because cluster.Synchronize can block) entities with synchronized access now have their own mutex.
@noxiouz @3Hren @oopcode PTAL


insonmnia/hub/acl.go
```diff
@@ -32,6 +35,34 @@ func NewACLStorage() ACLStorage {
...
+	return json.Marshal(set)
+}
+
+func (s *workerACLStorage) UnmarshalJSON(data []byte) error {
+	unmarshalled := make([]string, 0)
+	err := json.Unmarshal(data, &unmarshalled)
+	if err != nil {
+		return err
+	}
+	s.storage = set.NewNonTS()
```
3Hren: /sarcasm what if we have more than one thread?

insonmnia/hub/acl.go
```diff
@@ -3,6 +3,7 @@ package hub
 import (
 	"sync"
 
+	"encoding/json"
```
3Hren: Sort

insonmnia/hub/acl.go
```diff
@@ -32,6 +35,34 @@ func NewACLStorage() ACLStorage {
 	}
 }
 
+func (s *workerACLStorage) MarshalJSON() ([]byte, error) {
+	if s == nil {
+		return json.Marshal(nil)
+	}
+	s.mu.RLock()
```
3Hren: This will deadlock if is called under `Apply` (which is https://github.com/sonm-io/core/pull/182/files#diff-882aa38dff4c0888c4d85060cd652ac7L190).

3Hren:
Either I can't understand how it works for ACL set, or it will deadlock. Please explain the calling order for sync/fetch, it's not obvious.

insonmnia/hub/acl.go
antmat: What?

insonmnia/hub/server.go
```diff
@@ -1007,18 +1043,32 @@ func (h *Hub) processClusterEvent(value interface{}) {
...
+		defer h.slotsMu.Unlock()
+		h.slots = value
+	case ACLStorage:
+		h.acl = value
 	case error:
 		// continue reading from event channel as there could be pending events
 		go func() {
 			log.G(h.ctx).Warn("cluster failure, retrying after 10 seconds", zap.Error(value))
 			time.Sleep(time.Second * 10)
 			h.eventCh = h.cluster.Run()
```
noxiouz: is it safe to change h.eventCh w/o lock?

antmat: Reworked this part

insonmnia/hub/server.go
```diff
@@ -973,22 +997,39 @@ func (h *Hub) Serve() error {
...
+		log.G(h.ctx).Warn("cluster failure, retrying after 10 seconds", zap.Error(err))
+
+		t := time.NewTimer(time.Second * 10)
+		select {
+		case <-h.ctx.Done():
+			t.Stop()
+			return
+		case <-t.C:
+			t.Stop()
+			continue
```
noxiouz: no need  `continue`

util/waiter.go
```diff
@@ -0,0 +1,19 @@
+package util
+
+import "sync"
+
```
noxiouz: PTAL at https://godoc.org/golang.org/x/sync/errgroup

antmat: IMO, It has a bit different semantic

PR #183, opened 2017-11-17
feat: get devices by miner ID

oopcode:


insonmnia/hub/server.go
```diff
@@ -650,6 +628,29 @@ func (h *Hub) Devices(ctx context.Context, request *pb.Empty) (*pb.DevicesReply,
 	return reply, nil
 }
 
+func (h *Hub) MinerDevices(ctx context.Context, request *pb.ID) (*pb.DevicesReply, error) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
```
antmat: Looks like this lock is held a bit longer than needed.
Looks like h.getMinerByID fits here

oopcode: Fixed.

antmat:
Let's keep commit history clean. It' better to rebase on master to avoid merge commits.

Sokel:
uses merging instead of rebase due the merge conflicts, not did all.

PR #188, opened 2017-11-21
Fix: create deal while placing Marketplace's BID Order

sshaman1101:
This PR adds Deal creation when BID order is placing on the marketplace.

Added: 
- call blockchain API from the local node on order placement;
- load an Eth keys every "market" sub-command;

Changed: 
- OpenDeal() now receive pb.Deal model (instead of tons of func params)

### DIFF
--- a/insonmnia/node/market.go
+++ b/insonmnia/node/market.go
@@ -6,8 +6,11 @@ import (
 
 	"crypto/ecdsa"
 
+	"fmt"
+
 	log "github.com/noxiouz/zapctx/ctxlog"
 	"github.com/pkg/errors"
+	"github.com/sonm-io/core/blockchain"
 	"github.com/sonm-io/core/insonmnia/structs"
 	pb "github.com/sonm-io/core/proto"
 	"github.com/sonm-io/core/util"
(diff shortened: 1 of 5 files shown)


insonmnia/node/market.go
```diff
@@ -6,8 +6,11 @@ import (
 
 	"crypto/ecdsa"
 
+	"fmt"
```
noxiouz: no newline above

insonmnia/node/market.go
```diff
@@ -69,22 +72,30 @@ func HandlerStatusString(status uint8) string {
...
 }
 
 func newOrderHandler(ctx context.Context, loc string, o *pb.Order) (*orderHandler, error) {
 	ctx, cancel := context.WithCancel(ctx)
 
 	cc, err := util.MakeGrpcClient(ctx, loc, nil)
 	if err != nil {
-		log.G(ctx).Debug("cannot create locator client", zap.Error(err))
+		log.G(ctx).Error("cannot create locator client", zap.Error(err))
+		return nil, err
```
noxiouz: cancel() to prevent resource leak

insonmnia/node/market.go
```diff
@@ -181,9 +193,26 @@ func (h *orderHandler) propose(askID, supID string) error {
...
+		BuyerID:    order.GetByuerID(),
+		Price:      fmt.Sprintf("%d", order.Price),
+		Status:     pb.DealStatus_PENDING,
+		// TODO(sshaman1101): calculate hash
+		SpecificationHash: "",
+	}
+
+	_, err := h.bc.OpenDeal(key, deal)
+	if err != nil {
+		log.G(h.ctx).Debug("cannot open deal", zap.Error(err))
```
noxiouz: Debug?

PR #192, opened 2017-11-22
feat: worker authorization

3Hren:
At last workers ACL are used.

It's also possible to run hub/miner in insecure mode as before by setting GRPC_INSECURE enrironment variable.

There are four possible cases:
- Hub insecure, miner insecure -> everything as it was before.
- Hub secure, miner insecure -> hub rejects connection.
- Hub insecure, miner secure -> framing error while connecting.
- Hub secure, miner secure -> wallet check.

### DIFF
--- a/insonmnia/hub/miner_ctx.go
+++ b/insonmnia/hub/miner_ctx.go
@@ -2,6 +2,7 @@ package hub
 
 import (
 	"errors"
+	"fmt"
 	"net"
 	"sync"
 	"time"
@@ -24,6 +25,7 @@ var (
 	errOrderNotExists    = errors.New("specified order not exists")
 	errCPUNotEnough      = errors.New("number of CPU cores requested is unable to fit system's capabilities")
 	errMemoryNotEnough   = errors.New("number of memory requested is unable to fit system's capabilities")
+	errForbiddenMiner    = errors.New("miner is forbidden")
 )
 
 type OrderId string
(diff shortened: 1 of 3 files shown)


3Hren:
```
esafronov@5:~|⇒  python -c "import random; print(random.choice(['antmat', 'noxiouz', 'sshaman1101', 'oopcode']))"
antmat
```

insonmnia/hub/miner_ctx.go
```diff
@@ -58,16 +60,24 @@ type MinerCtx struct {
...
+		}
+	}
+
 	var (
 		m = MinerCtx{
 			conn:         conn,
 			statusMap:    make(map[string]*pb.TaskStatusReply),
 			usageMapping: make(map[OrderId]*resource.Resources),
 		}
-		err error
```
antmat: Looks like it's more convinient in this case to use named return values.

insonmnia/hub/miner_ctx.go
```diff
@@ -58,16 +60,24 @@ type MinerCtx struct {
...
 	)
 	m.ctx, m.cancel = context.WithCancel(ctx)
-	m.grpcConn, err = util.MakeGrpcClient(ctx, "miner", h.creds, grpc.WithDialer(func(_ string, _ time.Duration) (net.Conn, error) {
+	m.grpcConn, err = util.MakeGrpcClient(ctx, "miner", nil, grpc.WithDialer(func(_ string, _ time.Duration) (net.Conn, error) {
```
sshaman1101: Why `util.MakeGrpcClient` receive "miner" as address parameter?

antmat: Nothing will change if we replace "miner" with ""

PR #193, opened 2017-11-22
feat: worker-side Hub authorization

3Hren:
This commit activates worker-side Hub authorization using specified wallet key as a Hub endpoint. This wallet is compared with signed-one provided by a Hub during TLS handshake, verifying that we've connected to the proper Hub.

It is also planned to be used for discovery through the Locator, but for now a Hub address must also be specified. The result hub endpoint format is <ethAddr>@<socketAddr>.

This commit also makes hub settings required. I have no idea what to do with whisper at worker's side, but it seems that it can be safely removed.

### DIFF
--- a/insonmnia/miner/builder.go
+++ b/insonmnia/miner/builder.go
@@ -4,11 +4,10 @@ import (
 	"crypto/ecdsa"
 	"crypto/tls"
 	"fmt"
+	"net"
 	"os"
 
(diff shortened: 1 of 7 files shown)


3Hren:
```
python -c "import random; print(random.choice(['antmat', 'noxiouz', 'sshaman1101', 'oopcode']))"
noxiouz
```

PR #195, opened 2017-11-23
Feat/add events functions

Sokel:


blockchain/api.go
```diff
@@ -182,6 +194,168 @@ func (bch *api) CloseDeal(key *ecdsa.PrivateKey, id *big.Int) (*types.Transactio
 	return tx, err
 }
 
+func (bch *api) GetOpenedDeal(hubAddr *string, clientAddr *string) ([]*big.Int, error) {
```
sshaman1101: receive a string, not a pointer

blockchain/example/test_events.go
```diff
@@ -0,0 +1,31 @@
+package main
+
+import (
+	"log"
+
+	"github.com/sonm-io/core/blockchain"
+)
+
+func main() {
```
sshaman1101: It is not a test (like unit test), so remove a `test` word from a file name.
Just an `example/events.go` is OK, the name shows that this example file contains something about events.

blockchain/api.go
```diff
@@ -182,6 +194,168 @@ func (bch *api) CloseDeal(key *ecdsa.PrivateKey, id *big.Int) (*types.Transactio
...
+		}
+
+		out = append(out, l.Topics[3].Big())
+	}
+
+	// shift ids of opened deals by accepted and closed deals
+
+	return out, nil
+}
+
```
sshaman1101: So, you just copy-paste one piece of code three times in a row?

PR #199, opened 2017-11-23
marketplace and locator logging

sshaman1101:
This PR enables logging for Locator and Marketplace service.


insonmnia/marketplace/server_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"golang.org/x/net/context"
```
noxiouz: why not context from stdlib?

PR #200, opened 2017-11-23
fix: removed IP discovery through Amazon

oopcode:
This commit removes IP disovery through Amazon. Public IP discovery has changed.
* For Hub, public IP must be specified in configuration.
* For Miner, multiple IPs can be specified. If `firewall` settings are provided in configuration, a STUN server will used to discover IP. If there's no firewall settings and `public_ip_addrs` is provided in configuration, those addresses will be used. If none of the above holds, network interfaces are scanned for IPs. IPs found for miner are sorted as follows: first come all public IPs (IPv6 before IPv4), then all private IPs (IPv6 before IPv4).


insonmnia/hub/miner_ctx.go
```diff
@@ -332,15 +331,17 @@ func (m *MinerCtx) registerRoutes(ID string, ports map[string]*pb.SocketAddr) []
 			continue
 		}
 
-		route, err := m.router.RegisterRoute(ID, binding.Network(), endpoint.Addr, uint16(endpoint.Port))
```
oopcode: Please pay special attention to these changes.

insonmnia/hub/miner_ctx.go
```diff
@@ -332,15 +331,17 @@ func (m *MinerCtx) registerRoutes(ID string, ports map[string]*pb.SocketAddr) []
...
-		if err != nil {
-			log.G(m.ctx).Warn("failed to register route", zap.Error(err))
-			continue
+		for _, addr := range endpoint.Addrs {
+			route, err := m.router.RegisterRoute(ID, binding.Network(), addr, uint16(endpoint.Port))
```
3Hren: It's not ideal, because we actually can register all reals under the same VIP, but due to lack such API it's okay. I'll add it later.

insonmnia/miner/builder.go
```diff
@@ -89,37 +81,15 @@ func (b *MinerBuilder) Build() (miner *Miner, err error) {
...
-		log.G(b.ctx).Info("discovered public IP address",
-			zap.Any("addr", b.ip),
-			zap.Any("nat", b.nat),
-		)
+	publicIPs, err := b.getPublicIPs()
+	if err != nil {
+		return nil, err
 	}
 
+	log.G(b.ctx).Info("Discovered public IPs",
```
3Hren: s/D/d

insonmnia/miner/builder.go
```diff
@@ -225,6 +195,52 @@ func (b *MinerBuilder) Build() (miner *Miner, err error) {
 	pb.RegisterMinerServer(grpcServer, m)
 	return m, nil
 }
+
+func (b *MinerBuilder) getPublicIPs() (pubIPs []string, err error) {
+	// Discover IP if we're behind a NAT.
+	if b.cfg.Firewall() != nil {
+		log.G(b.ctx).Debug("Discovering public IP address with NAT type, this might be slow")
```
3Hren: s/D/d

proto/miner.proto
```diff
@@ -58,7 +58,7 @@ message MinerStartRequest{
 }
 
 message SocketAddr {
-    string addr = 1;
+    repeated string addrs = 1;
```
3Hren: It's not correct for a socket address to have multiple IPs. I'd rather return multiple SocketAddr instead.

util/util.go
```diff
@@ -157,3 +123,34 @@ func ParseTaskID(s string) (string, string, error) {
 
 	return id, hub, nil
 }
+
+func GetAddrInfo() (addrs []net.IP, err error) {
```
3Hren: I think we should rename it to avoid being confused with [getaddrinfo](http://man7.org/linux/man-pages/man3/getaddrinfo.3.html).

oopcode: I named it like that intentionally (because it does the same thing). But okay.

insonmnia/hub/miner_ctx.go
oopcode: Yes, there's no suitable API yet. We'll change that later.

3Hren:
LGTM

PR #203, opened 2017-11-25
feat: locator secure connections

oopcode:
* Locator now always uses secure TLS connections.
* Ethereum address for announcements is extracted from client's certificate.
* Configuration file is now required.


insonmnia/locator/config.go
```diff
@@ -0,0 +1,40 @@
...
+	return cfg, nil
+}
+
+func DefaultConfig(addr string) *LocatorConfig {
+	return &LocatorConfig{
+		ListenAddr:    addr,
+		NodeTTL:       time.Hour,
+		CleanupPeriod: time.Minute,
+		Eth: EthConfig{
+			PrivateKey: "d07fff36ef2c3d15144974c25d3f5c061ae830a81eefd44292588b3cea2c701c",
```
noxiouz: default for what? For tests?

oopcode: Not my code, but yes, it’s used in tests.

PR #206, opened 2017-11-27
Node: wait for approved deal

sshaman1101:
This PR fixes deal creation process into node;

Added intermediary status for order handler. Now the deal for an order must be approved on Hub side. Then deal will be marked as ready-to-use for the buyer.


3Hren:
What the hell with changes list?

sshaman1101:
It has happened because PRs #206 and #208 are merged in the wrong order. My bad.

PR #208, opened 2017-11-29
Hub and market integration

sshaman1101:
Chained with #206. This PR implements Ask order creation in Hub-side;

Added: 
- create ASK orders when a slot is inserted
- cancel ASK order when a slot is removed
- represent order price as a string, need to be converted to `*big.Int` when using Deals;
- store slots as a map, not a slice;


3Hren:
LGTM except goddammit todo's.

PR #215, opened 2017-11-30
grpc: redirected request brings proper wallet

noxiouz:


insonmnia/hub/server.go
```diff
@@ -184,7 +186,40 @@ func (h *Hub) onRequest(ctx context.Context, req interface{}, info *grpc.UnarySe
...
+		return nil, status.Error(codes.Unauthenticated, "no peer info")
+	}
+
+	switch au := p.AuthInfo.(type) {
+	case nil:
+		// GRPC_INSECURE should be set
+		return handler(ctx, req)
+	case util.EthAuthInfo:
+		if au.Wallet != h.ethAddr {
+			if h.acl.Has(au.Wallet) {
```
3Hren: Wait, I'm not sure whether we pass wallet with or without `0x` prefix.

noxiouz: I checked: both of them with 0x. Btw, seems we need special type for such things to hide format related stuff

Sokel: such as go-ethereum/common/Address?

noxiouz: yep, but our sturcts hold it as string

Sokel: typecast to common.Address and return one-formatted string already
see this: https://github.com/sonm-io/core/pull/237/files#diff-3ae177b094178b29d82a6311d08193a1R62

sshaman1101: In case of this, it may be better to move the `compareEthAddr` function to the util package.

noxiouz:
@Sokel @3Hren I changed wallet from string to ethereum.common.Address.

I gonna change all ethAddresses across the repo to this type

PR #218, opened 2017-12-04
feat: pull committed & completed image

3Hren:
### Added
 - `sonmcli task pull` command to pull committed container after it has been stopped. Requires user to be authenticated properly to prevent stealing foreign tasks. Can be piped with docker load by: `sonmcli task pull DEAL_ID CONTAINER_NAME TASK_ID | docker save`.

### Changed
- Forward commit-on-stop flag.

### Fixed
- Broken tests.


task.yaml
```diff
@@ -4,12 +4,10 @@ task:
   # docker image settings, required section
   container:
     # image name to start on miner, required param
-    name: image:v1
+    name: cocaine/cocaine-core:latest
```
sshaman1101: can you return the sample data back, please

3Hren: Ok, falling back to a non-working config.

cmd/cli/commands/tasks.go
```diff
@@ -193,6 +200,94 @@ var taskStartCmd = &cobra.Command{
...
+
+		var wr io.Writer
+		var err error
+		if taskPullOutput == "" {
+			wr = os.Stdout
+		} else {
+			wr, err = os.Create(taskPullOutput)
+			if err != nil {
+				return err
+			}
```
noxiouz: defer wr.Close()

insonmnia/miner/server.go
```diff
@@ -275,6 +275,35 @@ func (m *Miner) Load(stream pb.Miner_LoadServer) error {
...
 }
 
+func (m *Miner) Save(request *pb.SaveRequest, stream pb.Miner_SaveServer) error {
+	log.G(m.ctx).Info("handling Save request", zap.Any("request", request))
+
+	info, rd, err := m.ovs.Save(stream.Context(), request.ImageID)
+	if err != nil {
+		return err
+	}
+
```
noxiouz: defer rd.Close()

PR #224, opened 2017-12-04
fix: init market default client

sshaman1101:


antmat:
Please fix travis first

PR #233, opened 2017-12-05
Feat/extended task status

oopcode:
TaskStatusReply now contains information about available resources, including cgroups info.


insonmnia/miner/overseer.go
```diff
@@ -445,16 +447,30 @@ func (o *overseer) Start(ctx context.Context, description Description) (status c
 		// NOTE: I don't think it can fail
 		return
 	}
+
+	var cpuCount int
+	if description.Resources.CPUQuota > 0 {
+		cpuCount = int(description.Resources.CPUQuota / 100000)
```
3Hren: Quite strange conversion. It's impossible in the common case to convert CPU quota to cores number.

oopcode: It is based on this excerpt from Docker documentation:

screen shot 2017-12-07 at 14 27 13

However, I just checked that setting `--cpus=X.Y` on Linux sets `"NanoCpus"` to `X.Y * 10^9`and does not affect any other fields (at least on my machine). I'll add `"NanoCpus"` check.

PR #234, opened 2017-12-06
feat: default local keystore for node and cli

oopcode:
CLI and node  now use project-local keystore.


cmd/cli/config/config.go
```diff
@@ -60,6 +60,8 @@ func (cc *cliConfig) getConfigPath() (string, error) {
 func (cc *cliConfig) fillWithDefaults() {
 	cc.OutFormat = OutputModeSimple
 	cc.HubAddr = ""
+	cc.Keystore = "./keys"
+	cc.Passphrase = "any"
```
sshaman1101: I think it's not OK, because of CLI with default config (without a file on disk) must ask for passphrase interactively. So just leave this fields blank :)

PR #236, opened 2017-12-07
Fix/config unify

sshaman1101:
Chained with #235.

This PR makes the similar config directives similar.


3Hren:
Let's rename everything, part 2.

This will require cascading config updates for each testing member. I'd hold on this until at least end of testing period.

sshaman1101:
Each testing member now can be easily upgraded using the Ansible in less than 2 minutes. 
Do not see any problem with that.

3Hren:
Conflicts

PR #242, opened 2017-12-08
Feature/tokens approve

sshaman1101:
Chained with #240;

This PRs adds the `approve` method into CLI. This method allows user to set amount of tokens that can be managed by contract;


cmd/cli/commands/tokens.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/spf13/cobra"
+)
+
+var approveTokenCmd = &cobra.Command{
+	Use:    "approve <amount>",
+	Short:  "Approve tokens",
+	PreRun: loadKeyStoreWrapper,
+	Args:   cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		bch, err := blockchain.NewAPI(nil, nil)
```
noxiouz: Is it really needed two Nils?

sshaman1101: nils are for applying default params inside the blockchain API constructor. 

( this API constructor is made by @Sokel, seems it must be rewritten using `options`,  like gRPC do.)

Sokel: @noxiouz, @sshaman1101 good idea, thanks

PR #246, opened 2017-12-11
fix: better makefile install rule

antmat:
it's rather strange that make install builds insomnia, but do not build marketplace and locator.
This PR fixes this and failure in case of missing GOPATH var

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,11 @@ GOCMD=./cmd
 ifeq ($(GO), )
     GO=go
 endif
+
+ifeq ($(GOPATH), )
+    GOPATH=$(shell ls -d ~/go)
+endif
+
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -75,27 +80,10 @@ build/aux: build/locator build/marketplace
 
 build: build/insomnia build/aux
 
-install/miner: build/miner
+install: all
```
sshaman1101: So I think that the `install` target must require the `build` target, not the `all` one.

antmat: Why?

antmat: install is usually separated because it oftens require higher privilegies, but it's not this case and we don't need building and installing division, so I'd personally either depend install on all or vice versa

noxiouz: @antmat +1

PR #249, opened 2017-12-12
refactor: use ethcommon.Address instead of string for ethAddr

noxiouz:
As I mentioned in #215, it's better to use specific type for Addresses to prevent 0x hell


insonmnia/hub/eth.go
```diff
@@ -112,7 +112,8 @@ func (e *eth) CheckDealExists(id string) (bool, error) {
 		return false, err
 	}
 
-	idOK := deal.GetSupplierID() == util.PubKeyToAddr(e.key.PublicKey)
+	// NOTE: May GetSupplierID return common.Address?
+	idOK := deal.GetSupplierID() == util.PubKeyToAddr(e.key.PublicKey).Hex()
```
sshaman1101: `util.EqualAddresses()`?

noxiouz: See the Note. If GetSupplierID was common.Address

PR #252, opened 2017-12-12
fix: warn user if no hub connection from Node

sshaman1101:
Too boilerplate, but I cannot invent any more clean solution.


antmat:
And what does addr contain in this case? I think you can manually validate it.

sshaman1101:
The diff looks scary... It will be better to look at the whole [file](https://github.com/sonm-io/core/blob/10faad5dbd6ecac91afd6b0cd4b4417d6b41a2e4/insonmnia/node/hub.go).

@antmat @3Hren @noxiouz PTAL

insonmnia/node/hub.go
```diff
@@ -1,101 +1,91 @@
...
-	return h.hub.GetDeviceProperties(ctx, req)
+	return pb.NewHubClient(cc), nil
 }
 
-func (h *hubAPI) SetDeviceProperties(ctx context.Context, req *pb.SetDevicePropertiesRequest) (*pb.Empty, error) {
-	log.G(h.ctx).Info("handling SetDeviceProperties request")
-	return h.hub.SetDeviceProperties(ctx, req)
-}
+func (h *hubAPI) intercept(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	log.G(h.ctx).Debug("intercepting request", zap.String("method", info.FullMethod))
```
antmat: It was Info previously. Maybe better to leave 'handling blabla request'?

insonmnia/node/server.go
```diff
@@ -94,42 +94,40 @@ func New(ctx context.Context, c Config, key *ecdsa.PrivateKey) (*Node, error) {
...
 	}
-	pb.RegisterDealManagementServer(srv, deals)
-	log.G(ctx).Info("deals service registered")
 
 	tasks, err := newTasksAPI(opts)
 	if err != nil {
 		return nil, err
 	}
 
+	srv := util.MakeGrpcServer(creds, grpc.UnaryInterceptor(hub.(*hubAPI).intercept))
```
antmat: Looks rather strange that you use hubAPI interceptor for all services

cmd/cli/commands/common.go
```diff
@@ -51,6 +51,7 @@ var (
 
 	// errors
 	errCannotParsePropsFile = errors.New("cannot parse props file")
+	errNoHubConnection      = errors.New("cannot connect to hub via Node, please check Hub Address settings into Node's config")
```
antmat: Looks like you don't use it

insonmnia/node/server.go
sshaman1101: According to [docs](https://godoc.org/google.golang.org/grpc#UnaryInterceptor), it unable to register more that one UnaryInterceptor for the gPRC server. Or may I do it wrong and the handler needs to be registered in another way?

PR #253, opened 2017-12-13
Feature/node resume processing

sshaman1101:
this PR allows Node to continue processing orders after node restart. 

Changed: 
- Node can query Market for its own orders and can start processing after restart
- Market now able to search by supplier and buyer IDs
- Market now search by Order, not by Slot + OrderType


proto/marketplace.proto
```diff
@@ -17,12 +17,10 @@ service Market {
 }
 
 message GetOrdersRequest {
-    // Slot describe resuorces to search for
-    Slot slot = 1;
-    // OrderType describe a type of orders to search
-    OrderType orderType = 2;
+    // Order keeps slot and type for searching
+    Order order = 1;
```
antmat: It's a bad habbit - you are breaking API. It's ok for now, but unacceptable after release. PTAL at this https://developers.google.com/protocol-buffers/docs/gotutorial#extending-a-protocol-buffer

insonmnia/node/server.go
```diff
@@ -127,19 +128,26 @@ func New(ctx context.Context, c Config, key *ecdsa.PrivateKey) (*Node, error) {
...
+		orderProcessor: market.(*marketAPI).restartOrdersProcessing(),
 	}, nil
 }
 
 // Serve binds gRPC services and start it
 func (n *Node) Serve() error {
+	// start background services
+	err := n.orderProcessor()
+	if err != nil {
+		// is this must break start-up?
```
antmat: 's/is this must/should this/'

antmat: Should it?

proto/marketplace.proto
sshaman1101: I understand that this will break API, but for now, we have no stable API at all. 
After the release, we will change proto-things with care.

insonmnia/node/server.go
sshaman1101: Oh, sorry, I am not so smart.

PR #255, opened 2017-12-14
Feat/market and locator authentication

oopcode:
* Marketplace uses secure connections
* All components that communicate with Locator and Marketplace now check their ethereum addrs


cmd/marketplace/main.go
```diff
@@ -6,14 +6,15 @@ import (
...
+	"github.com/sonm-io/core/accounts"
 	"github.com/sonm-io/core/insonmnia/logging"
 	"github.com/sonm-io/core/insonmnia/marketplace"
 	"go.uber.org/zap"
 	"golang.org/x/net/context"
 )
 
 var (
-	listenAddr  = flag.String("addr", ":9095", "Marketplace service listen address")
+	configPath  = flag.String("config", "marketplace.yaml", "Path to locator config file")
```
antmat: copy pasted "locator"

insonmnia/marketplace/config.go
```diff
@@ -0,0 +1,23 @@
...
+	cfg := &MarketplaceConfig{}
+	err := configor.Load(cfg, path)
+	if err != nil {
+		return nil, err
+	}
+
+	return cfg, nil
+}
+
+type EthConfig struct {
```
antmat: How many Eth configs do we have in our repo?

cmd/marketplace/main.go
```diff
@@ -26,11 +27,29 @@ func main() {
...
 
+	ctx := context.Background()
+
+	cfg, err := marketplace.NewConfig(*configPath)
+	if err != nil {
+		log.GetLogger(ctx).Error("failed to load config", zap.Error(err))
+		os.Exit(1)
+	}
+
+	key, err := accounts.LoadKeys(cfg.Eth.Keystore, cfg.Eth.Passphrase)
```
antmat: Consider creating generic EthConfig and put processing inside

insonmnia/marketplace/config.go
oopcode: okay, will do!

cmd/marketplace/main.go
oopcode: done

cmd/cli/config/config.go
```diff
@@ -24,21 +24,20 @@ type Config interface {
...
+	Eth       accounts.EthConfig `yaml:"ethereum"`
+	OutFormat string             `required:"false" default:"" yaml:"output_format"`
 }
 
 func (cc *cliConfig) OutputFormat() string {
 	return cc.OutFormat
 }
 
 func (cc *cliConfig) PassPhrase() string {
-	return cc.Passphrase
```
antmat: I'd better use Eth directly or via method (e.g. Eth().Passphrase()), but ok for now

accounts/config.go
```diff
@@ -0,0 +1,19 @@
...
+	"crypto/ecdsa"
+)
+
+type EthConfig struct {
+	Passphrase string `required:"false" default:"" yaml:"pass_phrase"`
+	Keystore   string `required:"false" default:"" yaml:"key_store"`
+}
+
+func (c *EthConfig) LoadKey() (*ecdsa.PrivateKey, error) {
+	key, err := LoadKeys(c.Keystore, c.Passphrase)
```
antmat: Just a thought - I think we need to make some config validation and postprocessing, where we can load keys for example and make some usefull work (not now though)

PR #262, opened 2017-12-18
feat: miner resolves hub addr via locator

oopcode:
* If hub's address is just an ethereum addr, miner looks up hub's addr via locator.
* Addresses returned by locator are checked for reachability, the first reachable address is used.

### DIFF
--- a/insonmnia/miner/builder.go
+++ b/insonmnia/miner/builder.go
@@ -272,6 +272,10 @@ func (b *MinerBuilder) getHubConnectionInfo() (string, common.Address, error) {
 		return "", common.Address{}, fmt.Errorf("all hub endpoints are unreachable: %+v", encounteredErrors)
 	}
 
+	if _, _, err := net.SplitHostPort(hubSockAddr); err != nil {
+		return "", common.Address{}, err
+	}
+
 	return hubSockAddr, hubEthAddr, err
 }
 
--- a/insonmnia/miner/builder.go
+++ b/insonmnia/miner/builder.go
@@ -232,13 +232,18 @@ func (b *MinerBuilder) getPublicIPs() (pubIPs []string, err error) {
 
 func (b *MinerBuilder) getHubConnectionInfo() (string, common.Address, error) {
 	var (
-		hubEndpoint       = b.cfg.HubEndpoint()
-		encounteredErrors = make(map[string]error)
+		hubEndpoint                  = b.cfg.HubEndpoint()
+		encounteredErrors            = make(map[string]error)
+		hubSockAddr, hubEthAddr, err = util.ParseEndpoint(b.cfg.HubEndpoint())
 	)
-	if strings.Contains(hubEndpoint, "@") {
-		return util.ParseEndpoint(b.cfg.HubEndpoint())
-	} else {
-		resolved, err := b.locatorClient.Resolve(b.ctx, &pb.ResolveRequest{EthAddr: hubEndpoint})
+
+	if err != nil {
+		return "", common.Address{}, err
+	}
+
+	if strings.HasPrefix(hubSockAddr, ":") {
+		// Only hub's port is provided.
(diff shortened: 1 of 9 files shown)


oopcode:
```
$ python -c "import random; print(random.choice(['3Hren', '3Hren', '3Hren']))"
3Hren
```

insonmnia/miner/builder.go
```diff
@@ -223,15 +229,71 @@ func (b *MinerBuilder) getPublicIPs() (pubIPs []string, err error) {
...
+
+		log.G(b.ctx).Info("resolved hub endpoints", zap.Any("endpoints", resolved.IpAddr))
+
+		for _, addr := range resolved.IpAddr {
+			addr = strings.Replace(addr, "10001", "10002", 1)
+
+			log.G(b.ctx).Debug("trying hub endpoint", zap.Any("endpoint", addr))
+
+			dialer := net.Dialer{DualStack: true, Timeout: time.Second}
+			testCC, err := dialer.DialContext(b.ctx, "tcp", addr)
```
sshaman1101: defer `testCC.Close()`

insonmnia/miner/builder.go
```diff
@@ -223,15 +229,71 @@ func (b *MinerBuilder) getPublicIPs() (pubIPs []string, err error) {
...
+			if err != nil {
+				log.G(b.ctx).Debug(
+					"hub endpoint is unreachable", zap.Any("endpoint", addr),
+					zap.Any("error", err))
+			} else {
+				testCC.Close()
+				hubSockaddr, hubEthAddr, err := util.ParseEndpoint(hubEndpoint + "@" + addr)
+				if err == nil {
```
sshaman1101: what error will be returned if err is not nil?

insonmnia/miner/builder.go
oopcode: It is closed if error is not nil (see below); do we expect an open connection if there was an error?

insonmnia/miner/builder.go
oopcode: If `err` is not `nil` (which is sort of impossible in this case, but it's just a safety check) we move on to the next resolved address, and if none of the resolved addresses worked, we return an error.

oopcode: Oh, I get it, I'll make it return the most relevant error.

insonmnia/miner/builder.go
```diff
@@ -223,15 +229,71 @@ func (b *MinerBuilder) getPublicIPs() (pubIPs []string, err error) {
...
+func NewMinerBuilder(cfg Config, key *ecdsa.PrivateKey) (*MinerBuilder, error) {
+	b := &MinerBuilder{key: key}
 	b.Config(cfg)
-	return b
+
+	if b.ctx == nil {
+		b.ctx = context.Background()
+	}
+
+	_, TLSConf, err := util.NewHitlessCertRotator(b.ctx, b.key)
```
noxiouz: do not drop rotator. Its method Stop should be called

oopcode: It's actually dropped in a couple of other places around the project. Should I fix those spots?

insonmnia/miner/builder.go
```diff
@@ -223,15 +229,77 @@ func (b *MinerBuilder) getPublicIPs() (pubIPs []string, err error) {
...
+		resolved, err := b.locatorClient.Resolve(b.ctx, &pb.ResolveRequest{EthAddr: hubEndpoint})
+		if err != nil {
+			return "", common.Address{}, fmt.Errorf(
+				"failed to resolve hub addr from %s: %s", hubEndpoint, err)
+		}
+
+		log.G(b.ctx).Info("resolved hub endpoints", zap.Any("endpoints", resolved.IpAddr))
+
+		for _, addr := range resolved.IpAddr {
+			addr = strings.Replace(addr, "10001", "10002", 1)
```
3Hren: What the hell is it?

oopcode:
`grpc.Balancer` is scheduled for the next sprint (paired with miner builder rewrite).

PR #263, opened 2017-12-18
fix: possible Hub crashes on deal  accept error

sshaman1101:
This PR fixes possible crashes of the Hub when there are some problems with Deal accepting;


3Hren:
It'll be great if you write a test to see it actually crashes now.

PR #264, opened 2017-12-18
feat: entire task API authorization & authentication

3Hren:
The entire task API now requires wallet authentication.

When a deal is accepted a new auth rule is created at the Hub. For such deal only clients with proper auth info can start tasks, stop them etc.

Technically this is done using per-call authorization options, which requires wallet information passing through.

The wallet itself is self-signed using private key at the client side. The server can verify passed signed wallet to ensure that no one else will stop for example other tasks.


3Hren:
Upd: do not merge. I'll add tests in a couple of hours.

PR #269, opened 2017-12-19
feat: whitelist implementation

antmat:
PTAL

### DIFF
--- a/hub.yaml
+++ b/hub.yaml
@@ -69,3 +69,7 @@ locator:
 market:
   # marketplace gRPC endpoint, required
   endpoint: "8125721C2413d99a33E351e1F6Bb4e56b6b633FD@127.0.0.1:9095"
+
+whitelist:
+  url: "https://raw.githubusercontent.com/sonm-io/allowed-list/master/general_whitelist.json"
+  enabled: true
--- a/insonmnia/hub/server.go
+++ b/insonmnia/hub/server.go
@@ -7,11 +7,14 @@ import (
 	"io"
 	"math/rand"
 	"net"
+	"path/filepath"
 	"reflect"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/docker/distribution/reference"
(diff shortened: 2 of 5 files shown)


hub.yaml
```diff
@@ -69,3 +69,7 @@ locator:
 market:
   # marketplace gRPC endpoint, required
   endpoint: "8125721C2413d99a33E351e1F6Bb4e56b6b633FD@127.0.0.1:9095"
+
+whitelist:
+  url: "https://raw.githubusercontent.com/sonm-io/allowed-list/master/general_whitelist.json"
+  enabled: true
```
3Hren: seems redundant. we can keep whitelist as a pointer instead, it's optional anyway.

insonmnia/hub/server.go
```diff
@@ -476,6 +489,42 @@ func (h *Hub) startTask(ctx context.Context, request *structs.StartTaskRequest)
 	return reply, nil
 }
 
+func (h *Hub) checkImageWhitelist(registry string, image string, auth string) (bool, reference.Named, error) {
```
3Hren: I strictly against making the entire method for a single field `whitelist`. Decompose into `Whitelist` interface.

hub.yaml
antmat: As we discussed enabled whitelist is default

insonmnia/hub/whitelist.go
```diff
@@ -0,0 +1,109 @@
...
+		if err != nil {
+			return false, nil, err
+		}
+		allowed, err := w.digestAllowed(ref.Name(), (string)(digestedRef.Digest()))
+		return allowed, ref, err
+	}
+	dockerClient, err := dc.NewEnvClient()
+	if err != nil {
+		return false, nil, err
+	}
```
noxiouz: defer dockerClient.Close(). Otherwise, it can lead to leaking of idle http connections

PR #278, opened 2017-12-20
Fix/deal duration

sshaman1101:
This PR implements deal duration.

Changed:
- Slot now have duration (in seconds);
- When the Node is opening a deal it will add a correct order duration to a deal (and writing it to the blockchain).

Added: 
- the Order and the Slot now accept human-readable values for byte fields (RAM, storage, net in-out)

Fixed:
- accidentally fix the orderHandler processing interruption when the Node cannot resolve Hub IP via the Locator;


cmd/cli/task_config/slot.go
```diff
@@ -1,37 +1,36 @@
...
-
 type RatingConfig struct {
 	Buyer    int64 `yaml:"buyer" required:"true"`
 	Supplier int64 `yaml:"supplier" required:"true"`
 }
 
 type ResourcesConfig struct {
 	Cpu        uint64             `yaml:"cpu_cores" required:"true"`
-	Ram        uint64             `yaml:"ram_bytes" required:"true"`
+	Ram        string             `yaml:"ram_bytes" required:"true"`
```
antmat: Maybe next time better to make separate PR?

insonmnia/marketplace/server_test.go
```diff
@@ -258,6 +267,7 @@ func TestInMemOrderStorage_GetOrders_Count(t *testing.T) {
 			Price:     "1",
 			OrderType: pb.OrderType_BID,
 			Slot: &pb.Slot{
+				Duration:  uint64(structs.MinSlotDuration.Seconds()),
```
antmat: How many such slots do you have?
I wonder if you changed those with some kind of sed or did it by hand :)

insonmnia/node/market.go
```diff
@@ -322,12 +323,19 @@ func (m *marketAPI) CreateOrder(ctx context.Context, req *pb.Order) (*pb.Order,
 		return nil, errNotAnBidOrder
 	}
 
+	if time.Duration(req.GetSlot().GetDuration())*time.Second < structs.MinSlotDuration {
```
antmat: why don't you use structs.Slot?

insonmnia/marketplace/server_test.go
sshaman1101: It made by hand, of course!

insonmnia/node/market.go
sshaman1101: Hmm, seems reasonable, will do

3Hren:
Hub part: LGTM for now (after private discussion).

3Hren:
Voted for default slot for tests. The boilerplate gauge hit its historical value.

sshaman1101:
Make boilerplate ~~great again~~ reduction. @3Hren @antmat PTAL!

PR #282, opened 2017-12-20
fix: disable search by ANY order

sshaman1101:


antmat:
Please remind me, why do we need this?

PR #284, opened 2017-12-21
Feature: allow to explicitly set client to deal with

sshaman1101:
Chained with #278 

This PR allows setting buyer or seller ID to perform deal a with the already known client, directly.

Added:
- Accepting buyer and supplier IDs when creating orders on Marketplace.

TBD:
- [x] Send buyer or supplier ID to the Marketplace (both Node and Hub sides)
- [x] Node: if supplier ID is present in the CreateOrder request -> apply supplier ID as search param to marketplace
- [x] Hub: if buyer ID  is present in the ProposeDeal's ASK request -> check if caller has the same ID, else reject that proposal


insonmnia/hub/server.go
```diff
@@ -37,13 +37,11 @@ import (
...
 
 var (
-	ErrInvalidOrderType = status.Errorf(codes.InvalidArgument, "invalid order type")
-	ErrAskNotFound      = status.Errorf(codes.NotFound, "ask not found")
-	ErrDeviceNotFound   = status.Errorf(codes.NotFound, "device not found")
-	ErrMinerNotFound    = status.Errorf(codes.NotFound, "miner not found")
-	errDealNotFound     = status.Errorf(codes.NotFound, "deal not found")
-	errTaskNotFound     = status.Errorf(codes.NotFound, "task not found")
-	errImageForbidden   = status.Errorf(codes.PermissionDenied, "specified image is forbidden to run")
+	ErrDeviceNotFound = status.Errorf(codes.NotFound, "device not found")
```
oopcode: oh come on :(

sshaman1101: It's not mine! It's indentation via gofmt!
(BTW I do not remove this constant because of it still used somewhere in the tests)

insonmnia/hub/server.go
```diff
@@ -752,27 +750,45 @@ func (h *Hub) ProposeDeal(ctx context.Context, r *pb.DealRequest) (*pb.Empty, er
...
+	h.slotsMu.Unlock()
+
+	if !ok {
+		return nil, status.Errorf(codes.NotFound, "slot not found")
 	}
-	if !order.IsBid() {
-		return nil, ErrInvalidOrderType
+
+	bidOrder, err := h.market.GetOrderByID(h.ctx, &pb.ID{Id: r.GetBidId()})
+	if err != nil || bidOrder == nil {
```
oopcode: `bidOrder` may be `nil` while `err` is also `nil`?

PR #286, opened 2017-12-21
feat: synchronize deals into cluster

antmat:
subj
@noxiouz @sshaman1101 PTAL
Did not tested yet
Upd: tested minimal case


antmat:
Still need to sync DealCredentials(will make separate PR)

PR #281, opened 2017-12-20
WIP: feat: add OpenDealPending() func

Sokel:

PR #289, opened 2017-12-21
Fix: opened and approved deal selection

sshaman1101:
Chained with #278, #284.

This PR fixes an opened and an approved deal selection. 

Changed: 
- Using sync deal creation method (added in #281) to get a created deal ID on the Node;
- Reverse the deals traversal on Hub (newest first);


Sokel:
@3Hren review this please.

PR #294, opened 2017-12-22
refactor: deal lifecycle

3Hren:
Changed:
- Closing a deal using blockchain API when it expires at Hub-side.

Fixed:
- Releasing resources after deal closing/expiration.

Closes DEV-189, DEV-258.

### DIFF
--- a/insonmnia/structs/order.go
+++ b/insonmnia/structs/order.go
@@ -3,6 +3,7 @@ package structs
 import (
 	"errors"
 	"math/big"
+	"time"
 
 	pb "github.com/sonm-io/core/proto"
 	"github.com/sonm-io/core/util"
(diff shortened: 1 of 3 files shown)


insonmnia/structs/order.go
```diff
@@ -98,3 +99,7 @@ func (o *Order) GetType() pb.OrderType {
 func (o *Order) GetByuerID() string {
 	return o.inner.GetByuerID()
 }
+
+func (o *Order) GetDuration() time.Duration {
+	return time.Duration(o.inner.Slot.Duration)
```
sshaman1101: Inner slot keeps duration in seconds, so use it as `time.Duration(o.inner.Slot.Duration) * time.Second`

PR #295, opened 2017-12-23
Chore: rename miner to worker

sshaman1101:


3Hren:
Definitely not the hero of the day.

sshaman1101:
> Definitely not the hero of the day.

And now?

PR #354, opened 2017-12-26
chore: rename a type from 'Price' to the 'BigInt'.

sshaman1101:


proto/price.go
```diff
@@ -6,15 +6,15 @@ import (
...
+func NewPrice(v *big.Int) *BigInt {
+	return &BigInt{
 		Neg: v.Sign() < 0,
 		Abs: v.Bytes(),
 	}
 }
 
 // NewPriceFromString tries to construct a new price from the specified string.
-func NewPriceFromString(s string) (*Price, error) {
+func NewPriceFromString(s string) (*BigInt, error) {
```
3Hren: Function name.

proto/price.go
```diff
@@ -6,15 +6,15 @@ import (
 )
 
 // NewPrice constructs a new price using specified big.Int.
-func NewPrice(v *big.Int) *Price {
-	return &Price{
+func NewPrice(v *big.Int) *BigInt {
```
3Hren: Function name.

proto/price.go
```diff
@@ -24,7 +24,7 @@ func NewPriceFromString(s string) (*Price, error) {
 }
 
 // BigInt returns the current price as *big.Int.
-func (m *Price) BigInt() *big.Int {
+func (m *BigInt) BigInt() *big.Int {
```
3Hren: Something like `Unwrap` fits better here.

PR #355, opened 2017-12-26
Feature: improve dealing process

sshaman1101:
The first step to improve dealing process. 

Added:

Hub:
- the ApproveDeal method (a stub, currently);

Node:
- wait for approval via the Hub's method;
- small changes to the dealing flow;


insonmnia/hub/server.go
```diff
@@ -874,6 +874,11 @@ func (h *Hub) ProposeDeal(ctx context.Context, r *pb.DealRequest) (*pb.Empty, er
 	return &pb.Empty{}, nil
 }
 
+func (h *Hub) ApproveDeal(ctx context.Context, r *pb.DealRequest) (*pb.Empty, error) {
```
3Hren: Wrong argument type for this method: deal id is also required, while spec hash is unneeded.

But it's okay, because it is unimplemented now.

PR #358, opened 2017-12-27
chore: new miner constructor

oopcode:


3Hren:
Removed more than added! Keep it up!

PR #361, opened 2017-12-28
fix: CancelOrder now stops processing loop

oopcode:
N.B.: please review this carefully.


insonmnia/node/market.go
```diff
@@ -364,15 +351,21 @@ func (m *marketAPI) startExecOrderHandler(ord *pb.Order) {
 
 	for {
 		select {
-		// cancel context to stop polling for ordrs
+		// Cancel context to stop polling for orders.
```
noxiouz: out of scope, but defer tk.Stop()

PR #363, opened 2017-12-29
feat: improve router by allowing several reals

3Hren:
This PR improves router interface by allowing to register several real services under the same virtual service.

Thus, we're almost ready to perform L3 load balancing IPvN->IPvM routing in any combination.

### DIFF
--- a/insonmnia/hub/router_linux.go
+++ b/insonmnia/hub/router_linux.go
@@ -24,7 +24,7 @@ func newIPVSRouter(ctx context.Context, gate *gateway.Gateway, pool *gateway.Por
 	}
 }
 
-func (r *ipvsRouter) Register(ID string, protocol Protocol) (VirtualService, error) {
+func (r *ipvsRouter) Register(ID string, protocol string) (VirtualService, error) {
 	host, err := gateway.GetOutboundIP()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


insonmnia/hub/router_linux.go
```diff
@@ -37,7 +35,7 @@ func (r *ipvsRouter) RegisterRoute(ID string, protocol string, realIP string, re
 		return nil, err
 	}
 
-	serviceOptions, err := gateway.NewServiceOptions(host.String(), port, protocol)
+	serviceOptions, err := gateway.NewServiceOptions(host.String(), port, protocol.String())
```
noxiouz: Maybe it's better to change NewServiceOptions signature to accept a protocol as Protocol type instead of string?

insonmnia/hub/router_linux.go
```diff
@@ -136,3 +123,40 @@ func (r *ipvsRouter) Close() error {
...
+	vsID    string
+	options *gateway.ServiceOptions
+	gateway *gateway.Gateway
+}
+
+func (s *ipvsVirtualService) ID() string {
+	return s.vsID
+}
+
+func (s *ipvsVirtualService) AddReal(ID string, host string, port uint16) (*route, error) {
```
noxiouz: route is a private type but returned from a public method

insonmnia/hub/router_linux.go
3Hren: Actually introducing `Protocol` was really bad idea, because it gives no proc except type safety, while giving unnecessary complexity.
I'll revert these back.

insonmnia/hub/router_linux.go
3Hren: Fixed.

insonmnia/hub/router_linux.go
3Hren: Also I felt it out of the scope.

PR #364, opened 2017-12-30
some text corrections

xeaon:


3Hren:
Hi!

Thanks for the contribution! However, `Node`, `Hub`, `Orders` etc are commonly-used terms in the project, so we feel it more convenient to keep these in logs stated with a capital letter.

PR #376, opened 2018-01-10
refactor: do not use real CLI config path in the tests

sshaman1101:


cmd/cli/config/config_test.go
```diff
@@ -3,32 +3,26 @@ package config
...
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 const (
-	configDir  = ".sonm"
-	configName = "cli.yaml"
+	testConfigDir = "/tmp/_sonm"
```
3Hren: Better to generate temporary file using https://golang.org/pkg/io/ioutil/#TempFile. This is more stable and cross-platform.

PR #381, opened 2018-01-12
feat: only_public_ips mode for locator

oopcode:
Locator can now work in `only_public_ips` mode.

* If hub announces its IPs and none of them is public, an error is returned.
* If some of the announced IPs are public, announcement is successful (only public IPs are written to db, skipped IPs are logged).
* Config field `only_public_ips` is optional.

PR #382, opened 2018-01-13
fix: check if IP is loopback or unicast

sshaman1101:
This PR fixes an IP address checking from #381


3Hren:
Also check for link-local multicast.

PR #383, opened 2018-01-15
fix: protect handler's private fields with mutex and getters

sshaman1101:


insonmnia/node/market_test.go
```diff
@@ -153,11 +154,11 @@ func TestCreateOrder_FullAsyncOrderHandler(t *testing.T) {
 
 	// wait for async handler is finished
 	time.Sleep(1 * time.Second)
-	assert.True(t, len(inner.tasks) == 1, "Handler must not be removed")
+	assert.True(t, inner.countHandlers() == 1, "Handler must not be removed")
 
-	handlr := inner.tasks[created.Id]
-	assert.Equal(t, statusDone, handlr.status,
-		fmt.Sprintf("Wait for status %s, but has %s", statusMap[statusDone], statusMap[handlr.status]))
+	handlr, _ := inner.getHandler(created.Id)
```
3Hren: Here and below: seems wrong to ignore the second value. Use `require.*` instead here.

sshaman1101: Done!

PR #385, opened 2018-01-16
Fix/hw logs

sshaman1101:
Added: 
- log error if the worker cannot collect GPU info
- log collected hardware on worker's start

Fixed:
- handler error if NewMiner has failed

Outofscope:
- lowercase logs is awesome;
- do not clash variables and import names;

### DIFF
--- a/insonmnia/hardware/hardware.go
+++ b/insonmnia/hardware/hardware.go
@@ -1,13 +1,9 @@
 package hardware
 
 import (
-	"context"
-
-	log "github.com/noxiouz/zapctx/ctxlog"
 	"github.com/shirou/gopsutil/mem"
 	"github.com/sonm-io/core/insonmnia/hardware/cpu"
 	"github.com/sonm-io/core/insonmnia/hardware/gpu"
-	"go.uber.org/zap"
 )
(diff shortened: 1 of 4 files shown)


insonmnia/hardware/hardware.go
```diff
@@ -83,6 +88,7 @@ func (h *hardwareInfo) Info() (*Hardware, error) {
 
 	gpuInfo, err := h.GPU()
 	if err != nil {
+		log.G(h.ctx).Warn("cannot get GPU devices list", zap.Error(err))
```
3Hren: Return error instead.

insonmnia/hardware/hardware.go
```diff
@@ -56,6 +60,7 @@ type HardwareInfo interface {
 }
 
 type hardwareInfo struct {
+	ctx context.Context
```
3Hren: No need for this if we'll return on failed GPU fetching.

insonmnia/hardware/hardware.go
oopcode: fucking awesome yeah

PR #398, opened 2018-01-18
fix: show error about balance if there is no balance

sshaman1101:


sshaman1101:
@antmat You are right about the first item, that breaks the loop, thanks!

filter orders by balance and allowance to prevent dealing with too expensive orders is added.

PR #401, opened 2018-01-19
refactor: nat decomposition

3Hren:
Less files in the worker.


proto/nat.go
```diff
@@ -0,0 +1,28 @@
+package sonm
```
antmat: Are you sure it's ok to add custom code to proto folder(just curious - looks ok at first sight)?

proto/nat.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"github.com/ccding/go-stun/stun"
+)
+
+func NewNATType(nat stun.NATType) NATType {
+	switch nat {
+	case stun.NATNone:
+		return NATType_NONE
+	case stun.NATError, stun.NATUnknown:
```
antmat: minor: ordering is not the same

proto/nat.go
3Hren: It's okay.

PR #404, opened 2018-01-21
feat: customize CI

3Hren:
Added:
- Coverage

Changed:
- Parallelize builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,11 @@ language: "go"
 go:
     - 1.9.x
 
+env:
+  - BUILD_TYPE=Tests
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -21,5 +26,21 @@ install:
...
     - go get github.com/golang/mock/mockgen
-    - make all
-    - GO=$(which go) make deb
+    - |
+      case "${BUILD_TYPE}" in
+        Tests)
+          make all
+          ;;
+        Debian)
+          make all
```
antmat: Looks like we don't need it in debian build

PR #412, opened 2018-01-24
fix: decrease orders loading time

sshaman1101:
This PR fix deals retrieving for the `tasks list` command.

The deals used to obtain the Hub address, then resolving IP address via the Locator and then load tasks list via the gRPC.

the Node needs only accepted deals in that context because only with the accepted deal the Hub can start client's tasks.

the `tasks list` call now takes up to 2/3 less time from the previous implementation.


3Hren:
Ethereum-guy review is required.

Sokel:
it's just work properly.

PR #419, opened 2018-01-26
fix: create ovs and cancel func as late as possible

sshaman1101:
This PR defers an overseer creation to the end of Worker's init.
In that case, we don't need to call `cancel()` and `ovs.Close()` every time we met an error.

### DIFF
--- a/insonmnia/miner/server.go
+++ b/insonmnia/miner/server.go
@@ -169,7 +169,6 @@ func NewMiner(cfg Config, opts ...Option) (m *Miner, err error) {
 	if o.ovs == nil {
 		o.ovs, err = NewOverseer(ctx, cfg.GPU())
 		if err != nil {
-			cancel()
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


insonmnia/miner/server.go
```diff
@@ -106,81 +106,74 @@ func NewMiner(cfg Config, opts ...Option) (m *Miner, err error) {
...
 
 	log.G(o.ctx).Info("discovered public IPs",
 		zap.Any("public IPs", o.publicIPs),
 		zap.Any("nat", o.nat))
 
+	ctx, cancel := context.WithCancel(o.ctx)
+	if o.ovs == nil {
+		o.ovs, err = NewOverseer(ctx, cfg.GPU())
+		if err != nil {
+			cancel()
```
3Hren: Isn't it looks weird? Even if an overseer fails to construct itself why the Hell should we cancel its context?

PR #421, opened 2018-01-26
fefactor: remove pflag

sshaman1101:
This PR removes `pflag` library and use `cobra` instead.


cmd/cobra.go
```diff
@@ -0,0 +1,33 @@
...
+
+// Runner func starts service wrapper by Cobra command
+type Runner func()
+
+// NewCmd returns new cobra.Command with --config and --version flags attached and output set to os.Stdout
+func NewCmd(name, appVersion string, conf *string, version *bool, run Runner) *cobra.Command {
+	c := &cobra.Command{
+		Use: name,
+		Run: func(_ *cobra.Command, _ []string) {
+			if *version {
```
3Hren: I'd also check version != nil just for sanity.

PR #424, opened 2018-01-29
fix: BigInt serialization

sshaman1101:
This PR adds (un)marshal methods for the `proto.BigInt` type, fixes marshaling issues for the `pb.Order`'s price into the CLI.

Closes DEV-366.

### DIFF
--- a/proto/bigint.go
+++ b/proto/bigint.go
@@ -1,6 +1,7 @@
 package sonm
 
 import (
+	"encoding/json"
 	"fmt"
 	"math/big"
 )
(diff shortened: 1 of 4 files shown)


proto/bigint.go
```diff
@@ -44,3 +45,29 @@ func (m *BigInt) Unwrap() *big.Int {
 func (m *BigInt) Cmp(other *BigInt) int {
 	return m.Unwrap().Cmp(other.Unwrap())
 }
+
+func (m *BigInt) MarshalJSON() ([]byte, error) {
```
3Hren: What will be if we accept this by value?

PR #437, opened 2018-02-02
feat: monolithic hub state

oopcode:
Hub state is now dumped and loaded in one piece.

### DIFF
--- a/insonmnia/hub/state.go
+++ b/insonmnia/hub/state.go
@@ -1,0 +1,109 @@
+package hub
+
+import (
+	log "github.com/noxiouz/zapctx/ctxlog"
+	"github.com/sonm-io/core/insonmnia/resource"
+	"go.uber.org/zap"
+)
+
+type hubState struct {
+	Tasks    map[string]*TaskInfo
+	Devices  map[string]DeviceProperties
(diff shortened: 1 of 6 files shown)


insonmnia/hub/state.go
```diff
@@ -0,0 +1,109 @@
...
+	defer h.associatedHubsMu.Unlock()
+
+	h.devicePropertiesMu.Lock()
+	defer h.devicePropertiesMu.Unlock()
+
+	h.tasksMu.Lock()
+	defer h.tasksMu.Unlock()
+
+	state := &hubState{
+		Tasks:    h.tasks,
```
antmat: I bet you need deep copy here

insonmnia/hub/state.go
```diff
@@ -0,0 +1,109 @@
...
+	state := h.getState()
+
+	if err := h.cluster.Synchronize(state); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (h *Hub) loadState(state *hubState) error {
```
antmat: this is a method of Hub, why is it defined in other file? Again this is one vote for making a state a member of hub and defining methods of that state

PR #443, opened 2018-02-05
refactor: minify amount of libs used for GPU Tuners

sshaman1101:
This PR remove lots of GPU-related libraries collected on host, leaves an only minimal amount of libs required to set-up tasks executed on the graphics card.

NVidia:
- `libnvidia-ml.so` is required;
- leaving `nvidia-smi` as useful diagnostic tool;

Radeon:
- collecting libs on the host is not required at all, only the `/dev/dri` and `/dev/kfd` is required to be mounted into the container;

### DIFF
--- a/insonmnia/miner/gpu/radeon_tuner.go
+++ b/insonmnia/miner/gpu/radeon_tuner.go
@@ -4,22 +4,17 @@ package gpu
 
 import (
 	"context"
-	"net"
 	"os"
 
 	"github.com/docker/docker/api/types/container"
-	"github.com/docker/go-plugins-helpers/volume"
-	log "github.com/noxiouz/zapctx/ctxlog"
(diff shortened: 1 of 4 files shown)


insonmnia/miner/gpu/radeon_tuner.go
```diff
@@ -4,22 +4,17 @@ package gpu
...
-	"github.com/sshaman1101/nvidia-docker/nvidia"
-	"go.uber.org/zap"
 )
 
 type radeonTuner struct {
 	volumePluginHandler
 }
 
-func newRadeonTuner(ctx context.Context, opts ...Option) (Tuner, error) {
+func newRadeonTuner(_ context.Context, opts ...Option) (Tuner, error) {
```
3Hren: If the context doesn't used, maybe drop it?

PR #445, opened 2018-02-05
[WIP] feat: only leader serves clients

oopcode:
* Only leader's endpoints are announced;
* Forwarding to leader (which didn't work anyway) is removed, non-leaders return an error.


insonmnia/hub/cluster.go
```diff
@@ -43,14 +43,14 @@ type ClusterEvent interface{}
 // Specific type of cluster event emitted when new member joins cluster
 type NewMemberEvent struct {
 	endpointsInfo
-	Id string
```
oopcode: Go standard linter hates `Id`s and likes `ID`s, please do not resist.

PR #454, opened 2018-02-07
feature: present prices as Ethereum-compatible values

sshaman1101:
changed:
- allow a user to pass price in Ether units instead of Wei units via CLI;


3Hren:
Why math? There is no math there.

sshaman1101:
Moved to the `util` package.

sshaman1101:
It was done by accident, fixed!

PR #460, opened 2018-02-08
fix: listen for SIGTERM to perform correct termination

sshaman1101:


cmd/hub/main.go
```diff
@@ -57,7 +58,7 @@ func run() {
 
 	go func() {
 		c := make(chan os.Signal, 1)
-		signal.Notify(c, os.Interrupt)
+		signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
```
oopcode: What's the difference between using `os.Interrupt` and `syscall.SIGINT`?

sshaman1101: There is [no difference](https://github.com/golang/go/blob/master/src/os/exec_posix.go#L18).

PR #468, opened 2018-02-09
feat: monolithic hub state

oopcode:
* Hub state is now contained in a single synchronised entity;
* All maintenance is migrated to a single background loop.


insonmnia/auth/auth.go
```diff
@@ -24,9 +24,9 @@ func (e Event) String() string {
 // By default the router allows unregistered events, but this behavior can be
 // changed using `DenyUnregistered` option.
 type AuthRouter struct {
+	mu        sync.RWMutex
 	ctx       context.Context
 	log       *zap.Logger
-	mu        sync.RWMutex
```
3Hren: Leave out of the scope changes for another PR. Changing all the things everywhere can cause conflicts.

PR #486, opened 2018-02-17
fix cli install

Xoeseko:
When installing from repo the cli.yaml file fails to install hopefully this solves it.

### DIFF
--- a/debian/sonm-cli.install
+++ b/debian/sonm-cli.install
@@ -1,1 +1,2 @@
 sonmcli usr/bin
+$HOME/.sonm/cli.yaml


3Hren:
Hi!

Thanks for the contribution. Hovewer I doubt that your PR fixes something. The proper way is to teach the CLI to merge/override configuration files depending on global/local scope.

Also [here](https://github.com/sonm-io/core/blob/master/debian/rules#L17) is the place of where all configuration files are installed, and for the CLI it should be in the global scope (`/etc/sonm/...`).

PR #488, opened 2018-02-20
feat: NAT punching 

3Hren:
This PR adds the full NAT Punching Protocol implementation, allowing to connect Hub and its clients, including workers, even if they are located under a single or several NATs.
Also this commit adds a Rendezvous server, which acts as an intermediate metadata exchange point.

# Overview

There are several ways of implementing network address and port translation.
In our case the application (Hub, Worker, etc.) running on a node in the masqueraded network needs to determine the external address of the NAT, i.e., the address that its communication peers detect. This is done because it is desired to set up a direct communications path (either to save the cost of
taking the data via a server or to improve performance) between two clients both of which are behind separate NATs.

STUN classified NAT implementation as full-cone NAT, (address) restricted-cone NAT, port-restricted cone NAT or symmetric NAT and proposed a methodology for testing a device accordingly. However initially it was develop to transmit UDP packages, which is sadly not our case. Instead we borrowed its ideas and adapt them to be able to develop a new hand-made protocol specifically for establishing direct p2p TCP connections.

NPP is a protocol aimed to achieve mutual address resolution for both client and server. This is especially useful where there is no guarantee that both peers are reachable directly, i.e. behind a NAT for example.
The protocol allows for servers to publish their private network addresses while resolving their real remove address. This information is saved under some ID until a connection held and heartbeat frames transmitted.
When a client wants to perform a connection it goes to this server, informs about its own private network addresses and starting from this point a rendezvous can be achieved.
Both client and server are informed about remote peers public and private addresses and they may perform several attempts to create an actual p2p connection.

By trying to connect to private addresses they can reach each other if and only if they're both in the same LAN or VLAN.
If a peer doesn't have private address, the connection is almost guaranteed to be established directly from the private peer to the public one. If both of them are located under different NATs, a TCP punching attempt can be performed.
At last, when there is no hope, for example in the case of Symmetric NAT, a special relay server can be used to forward the traffic.

Relay servers are kind of services that act as a switch, waiting for two peers want to communicate with each other, and then route their traffic through itself. They should be used as a last resort in the case of over paranoid firewalls.

Servers should publish all of possibly reachable endpoints for all protocols they support. Its Ethereum ID is determined automatically through our peer authentication mechanism.
Clients should specify the desired protocol and ID for resolution. For both server and client their public endpoint is determined automatically.

Currently only TCP connection establishment is supported

# Technical details

TCP sockets usually have a one-to-one correspondence to TCP port numbers on the local host: after the application binds one socket to a particular local TCP port, attempts to bind a second socket to the same TCP port fail.
But for TCP hole punching to work, however, we need to use a single local TCP port to listen for incoming TCP connections and to initiate multiple outgoing TCP connections concurrently.

The implementation requires the underlying TCP/IP stack to support SO_REUSEADDR and SO_REUSEPORT options, because for successful NAT traversal it's crucial to reuse the same src ip:port to be routed through the same external ip:port each time.

Thus, the minimum linux kernel version is 3.9. About other platforms it's unclear.

The other hack is the simultaneous TCP open, which unfortunately requires tons of fds allocated on each connect attempt, so prepare to enlarge your open files limits.


insonmnia/npp/puncher.go
```diff
@@ -0,0 +1,302 @@
...
+
+	m.log.Debug("punching", zap.Any("addrs", addrs))
+
+	pending := make(chan connTuple)
+
+	// An example of clean architecture (TM) of how to gracefully close a listenerChannel.
+	// Nah, just kidding. Fuck go.
+	wg := &sync.WaitGroup{}
+	wg.Add(1 + len(addrs.PrivateAddrs))
+	go func() {
```
antmat: in general you don't need this. Just read (on line 239) from this channel (1 + len(addrs.PrivateAddrs)) times

insonmnia/npp/puncher.go
```diff
@@ -0,0 +1,302 @@
...
+		go func() {
+			conn, err := m.punchAddr(ctx, addrs.PublicAddr)
+			m.log.Info("using NAT", zap.Any("addr", addrs.PublicAddr), zap.Error(err))
+			pending <- newConnTuple(conn, err)
+			wg.Done()
+		}()
+	}
+
+	for _, addr := range addrs.PrivateAddrs {
+		go func() {
```
antmat: practically the same function as 10 lines above

insonmnia/npp/puncher.go
```diff
@@ -0,0 +1,302 @@
...
+			m.log.Info("using private address", zap.Any("addr", addr), zap.Error(err))
+			pending <- newConnTuple(conn, err)
+			wg.Done()
+		}()
+	}
+
+	waiter := make(chan connTuple)
+
+	go func() {
+		var errs []error
```
antmat: you don't use errs, are you sure you need this?

insonmnia/npp/puncher.go
```diff
@@ -0,0 +1,302 @@
...
+			}
+		}
+
+		if !connected {
+			waiter <- newConnTuple(nil, fmt.Errorf("failed to punch the network: all attempts has failed"))
+		}
+	}()
+
+	conn := <-waiter
+	return conn.unwrap()
```
antmat: rather big function, maybe decompose it?

insonmnia/rendezvous/server.go
```diff
@@ -0,0 +1,374 @@
...
+	if ok {
+		delete(candidates.servers, peer.ID)
+	}
+}
+
+func (m *Server) removeServerWatch(id string, peer Peer) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	candidates, ok := m.rv[id]
```
antmat: do you ever clean meetings?

3Hren: Yes.

PR #495, opened 2018-02-22
feat: generate grpc cmd client for any SONM service

sshaman1101:


cmd/cli/commands/cert.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func init() {
+	makeCertCmd.PersistentFlags().StringVar(&certCertOutput, "cert", "sonm.crt", "file to save certificate")
+	makeCertCmd.PersistentFlags().StringVar(&certKeyOutput, "key", "sonm.key", "file to save key")
+}
+
+var makeCertCmd = &cobra.Command{
+	Use:    "make-cert",
+	Short:  "Make TLS certificate for gRPC based on Eth key",
```
antmat: I think I've missed the point - why do we need this?

PR #496, opened 2018-02-22
feat: represent logging level as string

sshaman1101:
changed:
- Service configs now receive logging level as a string, not a numeric value. Fallback level is "debug".
- Config interface returns zapcore.Level to build logger instance from configured values.
- `debug` parameter is removed from BuildLogger, logger always in "production" mode with the eye-candy "console" encoding. Because nobody changes this mode during half a year `¯\(ツ)/¯`

### DIFF
--- a/hub.yaml
+++ b/hub.yaml
@@ -56,8 +56,8 @@ cluster:
 # Logging settings.
 logging:
   # The desired logging level.
-  # Allowed values in range of -1 (high verbosity) to 3 (most quiet)
-  level: -1
+  # Allowed values is "debug", "info", "warn", "error", "panic" and "fatal"
+  level: debug
 
 # blockchain-specific settings.
 ethereum:
--- a/hub.yaml
+++ b/hub.yaml
@@ -56,7 +56,7 @@ cluster:
 # Logging settings.
 logging:
   # The desired logging level.
-  # Allowed values is "debug", "info", "warn", "error", "panic" and "fatal"
+  # Allowed values are "debug", "info", "warn", "error", "panic" and "fatal"
   level: debug
 
 # blockchain-specific settings.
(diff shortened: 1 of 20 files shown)


hub.yaml
```diff
@@ -56,8 +56,8 @@ cluster:
 # Logging settings.
 logging:
   # The desired logging level.
-  # Allowed values in range of -1 (high verbosity) to 3 (most quiet)
-  level: -1
+  # Allowed values is "debug", "info", "warn", "error", "panic" and "fatal"
```
3Hren: is -> are

insonmnia/logging/logging_test.go
```diff
@@ -0,0 +1,28 @@
...
+	tests := []struct {
+		in  string
+		out zapcore.Level
+	}{
+		// any case must be parsed
+		{in: "warn", out: zapcore.WarnLevel},
+		{in: "Warn", out: zapcore.WarnLevel},
+		{in: "WARN", out: zapcore.WarnLevel},
+		// any other values must be represented as default debug level
+		{in: "-1", out: zapcore.DebugLevel},
```
3Hren: It is surprising for users. I'd rather throw an error on other values.

insonmnia/hub/config.go
```diff
@@ -67,6 +69,10 @@ type Config struct {
 	MetricsListenAddr string             `yaml:"metrics_listen_addr" default:"127.0.0.1:14000"`
 }
 
+func (c *Config) LogLevel() (zapcore.Level, error) {
+	return logging.ParseLogLevel(c.Logging.Level)
```
3Hren: Can't we parse this in `NewConfig` instead to produce errors in construction time instead of runtime? In some private field, which won't affect deserialization.
It saves some lines of code.

sshaman1101:
Desired log level now parsing when the config is loading, not when the value is extracted.
The `configor` package is updated because of [this bugfix](https://github.com/jinzhu/configor/pull/24) was merged.

PR #500, opened 2018-03-01
feat: store tinc plugin data in boltdb

antmat:
@oopcode @sshaman1101 @3Hren PTAL

### DIFF
--- a/insonmnia/miner/network/tinc_driver.go
+++ b/insonmnia/miner/network/tinc_driver.go
@@ -2,6 +2,7 @@ package network
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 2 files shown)


insonmnia/miner/network/tinc_driver.go
```diff
@@ -69,9 +90,58 @@ type TincNetworkState struct {
...
+			delErr := t.storage.Delete("state")
+			if delErr != nil {
+				t.logger.Errorf("could not cleanup storage for tinc network - %s", delErr)
+			}
+		}
+	}()
+
+	exists, err := t.storage.Exists("state")
+	if err != nil || !exists {
+		return err
```
sshaman1101: If err == nil && exists == false nil will be returned, but no state was loaded. The caller wouldn't be notified of an error.

insonmnia/miner/network/tinc_driver.go
```diff
@@ -69,9 +90,58 @@ type TincNetworkState struct {
...
+	}
+
+	data, err := t.storage.Get("state")
+	if err != nil {
+		return err
+	}
+
+	err = json.Unmarshal(data.Value, t)
+	for _, n := range t.Networks {
+		n.cli = t.cli
```
sshaman1101: What will be in `t.cli` if unmarshalling is failed? Looks like you can alter the structure internals without error checking.

insonmnia/miner/network/tinc_driver.go
```diff
@@ -69,9 +90,58 @@ type TincNetworkState struct {
...
+		n.logger = t.logger.With("source", "tinc/network/"+n.ID, "container", n.TincContainerID)
+	}
+	return err
+}
+
+func (t *TincNetworkState) sync() error {
+	var err error
+	defer func() {
+		if err != nil {
+			t.logger.Errorf("could not load sync network state - %s; erasing key", err)
```
sshaman1101: erasing key?

insonmnia/miner/network/tinc_driver.go
antmat: it's ok for us. No data - fresh start

insonmnia/miner/network/tinc_driver.go
antmat: In general that does not matter, but you are right - that would be useless

PR #504, opened 2018-03-05
chore: update configs and binaries location

sshaman1101:
Changed:
- Keep all configs into `etc`.
- Keep all binaries into `target`.
- Update paths into the Makefile.


sshaman1101:
OS type and arch added to the result binaries names. 

// cc @3Hren

PR #509, opened 2018-03-12
fix: pass context to the ethereum interface

sshaman1101:
We need to perform a correct cancellation and do not want to break a context when communicating with blockchain.


noxiouz:
+1

PR #510, opened 2018-03-13
ci: add target to preparing and uploading release binaries

sshaman1101:
added:
- Script for cross-compiling release binaries and upload it to the Github.
- Make target that calls the release script.

### DIFF
--- a/.ci/release.sh
+++ b/.ci/release.sh
@@ -1,0 +1,147 @@
+#!/bin/bash
+
+set -e
+
+VER=$1
+TOKEN=$2
+
+if [[ -z ${VER} ]]; then
+    echo "Err: please specify version to release"
+    exit 1
+fi
+
+if [[ -z ${TOKEN} ]]; then
+    echo "Err: please set Github token for release publishing"
+    exit 1
+fi
+
+
+LINUX_ZIP=sonm_${VER}_linux.zip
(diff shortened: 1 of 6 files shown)


3Hren:
Seems like a candidate for a separate repo.

.ci/release.sh
```diff
@@ -0,0 +1,147 @@
...
+    UPLOAD_DARWIN_REQ="${CURL} ${GH_AUTH} -H 'Content-Type: application/zip' --data-binary @'${DARWIN_ZIP_PATH}' ${UPLOAD_URL}?name=${DARWIN_ZIP}"
+
+    eval ${UPLOAD_LINUX_REQ}
+    eval ${UPLOAD_WINDOWS_REQ}
+    eval ${UPLOAD_DARWIN_REQ}
+
+    echo " >>> release uploading is complete";
+}
+remove_artifacts () {
+    rm -f sonmcli-darwin-10.6-amd64 sonmnode-darwin-10.6-amd64
```
3Hren: Looks like a hardcode.

sshaman1101: it's artifacts from xgo, file names always will be the same. But it will be better to replace with some wildcard statement. Will do.

sshaman1101:
> separate repo 

As we discussed with @abefimov separate repo for tools is OK.

PR #511, opened 2018-03-15
chore: update go-ethereum up to 1.8.2

sshaman1101:
Why?
`go-ethereum` now have the events filters (https://github.com/ethereum/go-ethereum/pull/15832), so we doesnt need to do not-so-smart polling from the blockchain.

As a next step, we need to rewrite some of `blockchain/api.go` internals and use new event filters.

Changed:
- A lot of new stuff is shipped into ./vendor.
- Minor fixes to match new go-ethereum API (gas price is represented as uint64, not big.Int).
- `insonmnia/npp/puncher.go` is updated because of vet is not happy (`addr` variable shadowing).


sshaman1101:
Closing because out of scope for now.

PR #513, opened 2018-03-15
feat: Searcher, Matcher and Dealer

sshaman1101:
Added:
- Searcher interface that hides Market/DWH communication (now for BIDs only).
- Matcher interface that picks up the best order. Split implementations for bid and ask.
- Dealer interface that handles dealing based on two orders. Now can perform full deal cycle from Node's side.

PR #515, opened 2018-03-16
refactor: use brand new Dealer into the Node

sshaman1101:
Chained with #513, so review it first!

changed:
- Using `Searcher`, `Selector` and `Dealer` interfaces into the Node.
- `OrderHandler` now keeps only an execution state and do not interact with remotes by itself.


insonmnia/dealer/match.go
```diff
@@ -0,0 +1,53 @@
...
+
+type askMatcher struct{}
+
+func NewAskMatcher() Matcher {
+	return &askMatcher{}
+}
+
+func (m *askMatcher) Match(orders []*sonm.Order) (*sonm.Order, error) {
+	if len(orders) == 0 {
+		return nil, nil
```
3Hren: I think we should return error in such case.

sshaman1101: Ouch, fixed this into #513

PR #530, opened 2018-03-23
feat: worker's benchmarks

sshaman1101:
This PR adds able to measure Worker's performance by executing
a bunch of benchmarks. Some of the benchmarks are present as statically-coded
functions that can simply pick-up a value from system (how many bytes of RAM
or CPU cores count). More complex measurement is implemented as Docker-containers
which must return a JSON string with the results.  It allows implementing
some amount of custom logic for each hardware group.

To keep the Worker's state, the boltdb was added as a simple storage.
The state is presented by three entities:

A list of benchmarks that were successfully executed for this Worker.
We need to keep this data between starts because benchmarking takes a lot of time, and we don't want to perform useless work at each restart.

A string with a hash calculated from hardware capabilities structure.
If hardware hash was changed we need to restart benchmarks and
measure newly installed hardware.

A mapping between hardware units and benchmark results. This data
is restored as internal Worker's state if hardware-hash is not changed and benchmark list is actual. Otherwise, we need to execute benchmark suite again and put actual results into the state storage.

For future digging and improvement:
- Drop unused .proto, remove deprecated methods.
- Calculate available memory in a right way (checking for parent cGroup settings).
- Calculate vmem amount for GPU devices.
- Fix calculation of GPU requirements for GPUProvider.


insonmnia/benchmarks/benchmarks.go
```diff
@@ -0,0 +1,51 @@
+package benchmarks
+
+import (
+	pb "github.com/sonm-io/core/proto"
+)
+
+const (
+	// we can get values for this benchmarks from the host system
+	CPUCores    = "cpu-cores"
```
3Hren: Use numeric ID instead.

insonmnia/hardware/hardware.go
```diff
@@ -1,78 +1,145 @@
...
+	GPU     []*GPUProperties   `json:"gpu"`
+	Memory  *MemoryProperties  `json:"memory"`
+	Network *NetworkProperties `json:"network"`
+	Storage *StorageProperties `json:"storage"`
+}
+
+// NewHardware returns initial hardware capabilities for Worker's host.
+// Parts of the struct may be filled later by HW-plugins.
+func NewHardware() (*Hardware, error) {
+	hw := &Hardware{
```
3Hren: It seems that the initialization is unnecessary.

insonmnia/miner/options.go
```diff
@@ -3,49 +3,24 @@ package miner
...
 	key       *ecdsa.PrivateKey
 	publicIPs []string
+	benchList benchmarks.BenchList
 }
 
 func (o *options) setupNetworkOptions(cfg Config) error {
 	var pubIPs []string
 
-	// Discover IP if we're behind a NAT.
-	if cfg.Firewall() != nil {
```
3Hren: Either return this or remove the `Firewall` method from the `Config`.

insonmnia/miner/server.go
```diff
@@ -716,11 +711,281 @@ func (m *Miner) TaskDetails(ctx context.Context, req *pb.ID) (*pb.TaskStatusRepl
...
+//
+// todo: test me
+func (m *Miner) isBenchmarkListMatches(required map[pb.DeviceType][]*pb.Benchmark, exiting map[string]bool) bool {
+	for _, benchs := range required {
+		for _, bench := range benchs {
+			if _, ok := exiting[bench.ID]; !ok {
+				return false
+			}
+		}
+
```
3Hren: Blank.

insonmnia/miner/server.go
```diff
@@ -716,11 +711,281 @@ func (m *Miner) TaskDetails(ctx context.Context, req *pb.ID) (*pb.TaskStatusRepl
...
+				c.Benchmark[bm.CPUCores] = &pb.Benchmark{
+					ID:     bm.CPUCores,
+					Result: uint64(c.Device.Cores),
+				}
+
+				continue
+			}
+
+			d := getDescriptionForBenchmark(ben)
+			d.Env["SONM_CPU_COUNT"] = fmt.Sprintf("%d", c.Device.Cores)
```
3Hren: Hardcode?

insonmnia/miner/server.go
```diff
@@ -716,11 +711,281 @@ func (m *Miner) TaskDetails(ctx context.Context, req *pb.ID) (*pb.TaskStatusRepl
...
+	case pb.DeviceType_DEV_NETWORK:
+		return m.runNetworkBenchGroup(benches)
+	case pb.DeviceType_DEV_STORAGE:
+		return m.runStorageBenchGroup(benches)
+	default:
+		return fmt.Errorf("unknown benchmark group \"%s\"", dev.String())
+	}
+}
+
+func (m *Miner) runCpuBenchGroup(benches []*pb.Benchmark) error {
```
3Hren: s/Cpu/CPU

insonmnia/miner/server.go
```diff
@@ -716,11 +711,281 @@ func (m *Miner) TaskDetails(ctx context.Context, req *pb.ID) (*pb.TaskStatusRepl
...
+
+			// save benchmark resutls for current CPU unit
+			c.Benchmark[ben.GetID()] = &pb.Benchmark{ID: ben.GetID(), Result: res.Result}
+		}
+	}
+
+	return nil
+}
+
+func (m *Miner) runRamBenchGroup(benches []*pb.Benchmark) error {
```
3Hren: s/Ram/RAM

insonmnia/miner/server.go
```diff
@@ -716,11 +711,281 @@ func (m *Miner) TaskDetails(ctx context.Context, req *pb.ID) (*pb.TaskStatusRepl
...
+		}
+
+		m.hardware.Memory.Benchmark[ben.GetID()] = &pb.Benchmark{ID: ben.GetID(), Result: res.Result}
+		return nil
+	}
+
+	return nil
+}
+
+func (m *Miner) runGpuBenchGroup(benches []*pb.Benchmark) error {
```
3Hren: s/Gpu/GPU

insonmnia/miner/server.go
```diff
@@ -716,11 +711,281 @@ func (m *Miner) TaskDetails(ctx context.Context, req *pb.ID) (*pb.TaskStatusRepl
...
+	}
+
+	return v.Results, nil
+}
+
+func getDescriptionForBenchmark(b *pb.Benchmark) Description {
+	return Description{
+		Image: b.GetImage(),
+		Env: map[string]string{
+			"SONM_BENCHMARK_ID": b.GetID(),
```
3Hren: Hardcore.

proto/benchmarks.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+
+package sonm;
+
+//  BenchmarkType describes hardware group for which this benchmark is applicable
+enum DeviceType {
+    DEV_UNKNOWN = 0;
```
3Hren: Do we need these prefixes?

PR #535, opened 2018-03-27
feat: REST interface for node

antmat:
Rest interface for Node.
@3Hren @sshaman1101 @oopcode PTAL

### DIFF
--- a/insonmnia/node/server.go
+++ b/insonmnia/node/server.go
@@ -2,6 +2,7 @@ package node
 
 import (
 	"crypto/ecdsa"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"io"
@@ -13,6 +14,7 @@ import (
 	"github.com/sonm-io/core/blockchain"
 	pb "github.com/sonm-io/core/proto"
 	"github.com/sonm-io/core/util"
+	"github.com/sonm-io/core/util/rest"
 	"github.com/sonm-io/core/util/xgrpc"
 	"go.uber.org/zap"
 	"golang.org/x/net/context"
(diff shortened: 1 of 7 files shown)


insonmnia/node/server.go
```diff
@@ -78,20 +80,29 @@ func newRemoteOptions(ctx context.Context, key *ecdsa.PrivateKey, conf Config, c
...
+	ctx     context.Context
+	cancel  context.CancelFunc
+	privKey *ecdsa.PrivateKey
 }
 
 // New creates new Local Node instance
 // also method starts internal gRPC client connections
 // to the external services like Market and Hub
 func New(ctx context.Context, c Config, key *ecdsa.PrivateKey) (*Node, error) {
+	ctx, cancel := context.WithCancel(ctx)
```
noxiouz: defer cancel()

util/rest/aes.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &AESDecoderEncoder{
+		cipherBlock: c,
+	}, nil
+}
+
+func (d *AESDecoderEncoder) DecodeBody(request *http.Request) (io.Reader, error) {
+	data, err := ioutil.ReadAll(request.Body)
```
noxiouz: defer request.Close()

antmat: From docs:
	// For server requests the Request Body is always non-nil
	// but will return EOF immediately when no body is present.
	// The Server will close the request body. The ServeHTTP
	// Handler does not need to.

insonmnia/node/server.go
```diff
@@ -78,20 +80,29 @@ func newRemoteOptions(ctx context.Context, key *ecdsa.PrivateKey, conf Config, c
...
+	httpSrv *rest.Server
+	srv     *grpc.Server
+
+	// services, responsible for request handling
+	hub    pb.HubManagementServer
+	market pb.MarketServer
+	deals  pb.DealManagementServer
+	tasks  pb.TaskManagementServer
+
+	//
```
3Hren: What?

PR #536, opened 2018-03-27
refactor: brand new Node API

sshaman1101:
added:
- `masterManagement` API (name may be changed into the future) that manage Worker-Master relations.

changed:
- `hubManagement` API is renamed to `workerManagement` API and interact with the Worker instances.
- ACL and device properties API methods where removed.
- Worker's device and tasks management APIs are squashed into one method for each API.

TBD:
- Split and simplify the market API exposed by the Node, improve the `Processing` method.
- Implement `Tasks` and `Devices` method with new API (blocked by #530).
- Implement `masterManagement` API when DWH becomes ready.


proto/marketplace.proto
```diff
@@ -7,14 +7,20 @@ import "timestamp.proto";
 package sonm;
 
 service Market {
+    //
```
antmat: :grey_question:

insonmnia/node/master.go
```diff
@@ -0,0 +1,35 @@
...
+func newMasterManagementAPI(opts *remoteOptions) sonm.MasterManagementServer {
+	return &masterMgmtAPI{
+		ctx:     opts.ctx,
+		remotes: opts,
+	}
+}
+
+func (m *masterMgmtAPI) WorkersList(context.Context, *sonm.Empty) (*sonm.Empty, error) {
+	ctxlog.G(m.ctx).Info("handling WorkersList request")
+	return &sonm.Empty{}, nil
```
antmat: return error in stubs

antmat: reserve a return message for this method

proto/node.proto
```diff
@@ -69,36 +69,23 @@ message DealStatusReply {
...
-    // WorkersList prouces a list of connected Workers
-    rpc WorkersList (Empty) returns (ListReply) {}
-    // WorkersStatus produces a detailed info about a Worker with given ID
-    rpc WorkerStatus (ID) returns (InfoReply) {}
+    // rpc DeviceList(Empty) returns (DevicesReply) {}
+    // GetDeviceInfo provides detailed information about device
+    // and show benchmark results.
+    //
+    // TODO: returns smth that can represent miner.hardware (extend Capabilities?)
+    rpc GetDeviceInfo(Empty) returns (DevicesReply) {}
```
antmat: Just DeviceInfo or Devices

antmat: btw, previous DeviceList was good enough

PR #548, opened 2018-03-29
feat: Introducing DWH (antmat edition)

oopcode:
This PR introduces DWH (a gRPC-based fast cache for new market data).


blockchain/util.go
```diff
@@ -41,3 +50,24 @@ func getCallOptions(ctx context.Context) *bind.CallOpts {
...
+}
+
+func extractBig(log types.Log, pos int) (*big.Int, error) {
+	if len(log.Topics) < pos {
+		return nil, errors.New("topic index out of range")
+	}
+
+	id, ok := big.NewInt(0x0).SetString(log.Topics[pos].Hex(), 0)
+	if !ok {
+		return nil, errors.Errorf("failed to parse to *big.Int: %+v", log.Topics[1].Hex())
```
sshaman1101: `log.Topics[1].Hex()` index is hardcoded?

etc/dwh.yaml
```diff
@@ -0,0 +1,22 @@
...
+ethereum:
+  # path to keystore
+  key_store: "./keys"
+  # passphrase for keystore
+  pass_phrase: "any"
+
+
+blockchain:
+  eth_endpoint: "0x0"
+  gas_price: 10
```
sshaman1101: I think this setting is bit meaningless. Gas price in our private blockchain is always 0 (as @Sokel says).
`eth_endpoint` is a contract address? I think endpoint must be presented by something like URL. We already have some options into `blockchain/options.go`.

PR #549, opened 2018-03-29
feat: read memory limit from CGroup

sshaman1101:
Previously we used total amount RAM on the host system as an available memory for scheduling.
But resources for containers may be limited by a parent CGroup settings.

This PR changes how available memory amount is measured.
The lesser value of `total RAM on host, a limit set for CGroup`
is used.

Changed:
- CGroups management code moved into a separated package to prevent cycle imports issue.
- `insonmnia/hardware/mem` package was updated to unify the hardware descriptions.
- CGroup interface extended with `Stats` method, used to get an internal cgroup stats for various resources.

### DIFF
--- a/insonmnia/cgroups/cgroup_linux.go
+++ b/insonmnia/cgroups/cgroup_linux.go
@@ -1,6 +1,6 @@
 // +build linux
 
-package miner
+package cgroups
 
 import (
 	"fmt"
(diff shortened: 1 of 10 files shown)


3Hren:
Description?

insonmnia/cgroups/cgroup_linux.go
```diff
@@ -20,7 +20,17 @@ type cgroup struct {
...
+	cgStat, err := c.Stat(cgroups.IgnoreNotExist)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Stats{MemoryLimit: cgStat.Memory.HierarchicalMemoryLimit}, nil
+}
+
+func (c *cgroup) New(name string, resources *specs.LinuxResources) (CGroup, error) {
+	resources.Memory.Swap = new(int64)
```
antmat: as requested from @sshaman1101 : "ti dolboeb"

PR #559, opened 2018-04-01
fix: tune benchmark list loading

sshaman1101:
Changed:
- Downloading of the benchmarks list can be skipped for development purposes.
- List now can be loaded from the filesystem, the `file://` schema support added.

### DIFF
--- a/etc/worker.yaml
+++ b/etc/worker.yaml
@@ -58,6 +58,7 @@ store:
   path: "/var/lib/sonm/worker.boltdb"
 
 benchmarks:
+  # URL to download benchmark list, use `file://` schema to load file from a filesystem.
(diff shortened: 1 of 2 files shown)


etc/worker.yaml
```diff
@@ -58,7 +58,10 @@ store:
   path: "/var/lib/sonm/worker.boltdb"
 
 benchmarks:
+  # URL to download benchmark list, use `file://` schema to load file from a filesystem.
   url: "https://raw.githubusercontent.com/sonm-io/allowed-list/master/benchmarks_list.json"
+  # downloading can be skipped for development purposes
+  # disable: true
```
3Hren: Better not to specify the URL at all instead of this. So make both `url` and  `benchmarks` (for now) optional.

sshaman1101: Done.

PR #580, opened 2018-04-06
feat: relay metrics

3Hren:
This commit introduces Relay metrics.

These are collected on relay servers, including current connections, tx/rx bytes transmitted from/into a named server.

Metrics can be obtained using embedded gRPC monitoring server.

Review #579 at first.


insonmnia/npp/relay/server.go
```diff
@@ -134,6 +237,52 @@ func (m *connPool) popRandom() *meeting {
...
+	})
+
+	return wg.Wait()
+}
+
+func (m *meetingHandler) transmitTCP(from, to net.Conn, metrics *atomic.Uint64, log *zap.SugaredLogger) error {
+	buf := make([]byte, m.bufferSize)
+
+	for {
+		bytesRead, err := from.Read(buf[:])
```
noxiouz: AFAIK, bytesRead can be non-zero even in case non-nil err

3Hren: Fixed in #582, because it's out of the scope (just code moving here).

PR #595, opened 2018-04-11
refact: use new AskPlan in worker API instead of slot

antmat:
Use new proto AskPlan in worker and node, instead of old combination of slot and price
@sshaman1101 @3Hren PTAL


cmd/cli/commands/printers.go
```diff
@@ -257,24 +257,21 @@ func printProcessingOrders(cmd *cobra.Command, tasks *pb.GetProcessingReply) {
...
-		if len(slots) == 0 {
+		plans := slots.GetAskPlans()
+		if len(plans) == 0 {
 			cmd.Printf("No Ask Order configured\r\n")
 			return
 		}
 
-		for id, slot := range slots {
+		for id, plan := range plans {
+			gpuDevices, _ := json.Marshal(plan.Resources.GetGPU().GetDevices())
```
sshaman1101: `strings.Join` ?

PR #607, opened 2018-04-17
feat: create order from bid.yaml

sshaman1101:
This PR adds able to read BID orders from the yaml spec
and place them on Marketplace.

Changed:
- Node's market API accepts `BidOrder` message, not `Order` itself. It allows providing more compact and pretty interface. Also, it prevents us from downloading and parsing benchmarks list on client-side.
- Now the Node able to obtain the benchmarks list. It using to validate `BidOrder`'s benchmarks by their codes.
- `BenchmarkList` interface now have two methods to provide the same list into different usable formats.


sshaman1101:
Blockchain API was fixed, so code into this PR becomes working.
@antmat @3Hren @oopcode PTAL!

insonmnia/benchmarks/benchmarks.go
```diff
@@ -27,11 +27,13 @@ const (
 )
 
 type BenchList interface {
-	List() map[pb.DeviceType][]*pb.Benchmark
+	MapByType() map[pb.DeviceType][]*pb.Benchmark
```
3Hren: The name is ambiguous.

proto/bid.proto
```diff
@@ -1,29 +0,0 @@
-syntax = "proto3";
```
3Hren: What's the reason to do this?

sshaman1101: Because of `bid.proto` imports `marketplace.proto` to re-use some types, and `marketplace.proto` imports `bid.proto` to use `BidOrder` message as parameter for `CreateOrder` method.

PR #623, opened 2018-04-20
chore: yet another protobuf cleanup

sshaman1101:
Removed:
- Previous resources descriptions (network capabilities, GPU count).
- Debris of the Miner-Hub protocol (Ping, Discover).
- CQRS-backed marketplace heritage (GetProcessingReply, TouchOrdersRequest).


proto/node.proto
```diff
@@ -15,7 +15,7 @@ package sonm;
 // to manage tasks running into the SONM network
 service TaskManagement {
     // List produces a list of all tasks running on different SONM nodes
-    rpc List(TaskListRequest) returns (TaskListReply) {}
+    rpc List(ID) returns (TaskListReply) {}
```
antmat: Maybe EthAddr4ess is better here?

proto/insonmnia.proto
```diff
@@ -132,18 +110,11 @@ message TaskLogsChunk {
...
-}
-
 message TaskResourceRequirements {
     // CPUCores specifies the number of CPU cores required.
     uint64 CPUCores = 1;
     // MaxMemory specifies the maximum memory in bytes required.
     int64 maxMemory = 2;
-    // GPUCount Describes whether a task requires GPU support.
-    // todo: completely useless in that representation.
-    GPUCount GPUSupport = 3;
```
antmat: Looks like it should be changed to something else instead of deletion, but ok for now

proto/deal.proto
```diff
@@ -5,15 +5,15 @@ package sonm;
 import "bigint.proto";
 import "timestamp.proto";
 
-// TODO: deprecated.
```
antmat: Why don't you like TODOs?

insonmnia/structs/resources.go
```diff
@@ -28,19 +28,10 @@ func NewTaskResources(r *pb.TaskResourceRequirements) (*TaskResources, error) {
...
-	case pb.GPUCount_SINGLE_GPU:
-		numGPUs = 1
-	default:
-	}
-
 	return resource.Resources{
 		NumCPUs: int(r.inner.GetCPUCores()),
 		Memory:  r.inner.GetMaxMemory(),
-		NumGPUs: numGPUs,
+		NumGPUs: -1,
```
antmat: How about removing NumGPUs from resource.Resources?

sshaman1101: `NumGPUs` field still used somewhere into the old resource scheduler code, seems out of the scope for now.

PR #628, opened 2018-04-22
feat: finalising DWH API

oopcode:
* Implemented remaining API methods
* Silenced internal errors for API methods
* More tests, refactoring


insonmnia/dwh/server.go
```diff
@@ -387,24 +542,41 @@ func (w *DWH) getProfileInfo(ctx context.Context, request *pb.ID) (*pb.Profile,
...
 	w.mu.RLock()
 	defer w.mu.RUnlock()
 
-	rows, err := w.db.Query(w.commands["selectBlacklists"], adderID.Id)
+	return w.getBlacklist(ctx, request)
+}
+
+func (w *DWH) getBlacklist(ctx context.Context, request *pb.BlacklistRequest) (*pb.BlacklistReply, error) {
+	var filters []*filter
+	if request.OwnerID > "0" {
```
sshaman1101: why "0" ? What if ownerID is an empty string? Maybe try to typecast into `common.address`?

oopcode: If `OwnerID` is an empty string, it will be less than `"0"`.

PR #633, opened 2018-04-23
dwh: check rows.Err()

noxiouz:


oopcode:
Can't seem to find anything related to missed `rows.Close()` in this diff!
Everything else is cool.

PR #647, opened 2018-04-25
feat: dwh event fallback and proper contexts

oopcode:
* retry for failed events
* proper context handling

### DIFF
--- a/cmd/dwh/main.go
+++ b/cmd/dwh/main.go
@@ -3,6 +3,8 @@ package main
 import (
 	"context"
 	"os"
+	"os/signal"
+	"syscall"
 
 	log "github.com/noxiouz/zapctx/ctxlog"
 	"github.com/sonm-io/core/cmd"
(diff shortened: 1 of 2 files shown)


cmd/dwh/main.go
```diff
@@ -40,7 +40,7 @@ func run() {
 
 	w, err := dwh.NewDWH(ctx, cfg, key)
 	if err != nil {
-		log.G(ctx).Error("cannot start DWH service", zap.Error(err))
+		log.G(ctx).Error("cannot create NewDWH", zap.Error(err))
```
antmat: cannot create NewDWH ?

cmd/dwh/main.go
```diff
@@ -49,7 +49,7 @@ func run() {
 	log.G(ctx).Info("starting DWH service", zap.String("grpc_addr", cfg.GRPCListenAddr),
 		zap.String("http_addr", cfg.HTTPListenAddr))
 	if err := w.Serve(); err != nil {
-		log.G(ctx).Error("cannot start DWH service", zap.Error(err))
+		log.G(ctx).Error("DWH failed", zap.Error(err))
```
antmat: It will be smth like "DWH failed - context canceled" during shutdown, won't it?

insonmnia/dwh/server.go
```diff
@@ -820,26 +817,55 @@ func (w *DWH) watchMarketEvents() error {
...
-	}
+			// Events in the same block can come in arbitrary order. If two events have to be processed
+			// in a specific order (e.g., OrderPlaced > DealOpened), we need to retry if the order is
+			// messed up.
+			if err := w.processEvent(event); err != nil {
+				w.logger.Error("failed to processEvent, retrying", zap.Error(err),
+					zap.Uint64("block_number", event.BlockNumber),
+					zap.String("event_type", reflect.TypeOf(event.Data).Name()))
+
+				go func() {
```
antmat: better extract method here

cmd/dwh/main.go
oopcode: "context cancelled" is already present in the error itself:

```
// Canceled is the error returned by Context.Err when the context is canceled.
var Canceled = errors.New("context canceled")
```

cmd/dwh/main.go
oopcode: you mean, missing space?

PR #666, opened 2018-04-28
feat: adapt market to worker

antmat:
Ready for review.
This PR introduces new resource pool, responsible for resources consuming/releasing and a salesman entity, which is responsible for all interaction with market and selling resources according to created ask plans.


insonmnia/miner/salesman.go
```diff
@@ -0,0 +1,343 @@
...
+
+func NewSalesman(
+	ctx context.Context,
+	state *state.Storage,
+	resources *resource.Scheduler,
+	hardware *hardware.Hardware,
+	eth blockchain.API,
+	matcher matcher.Matcher,
+	ethkey *ecdsa.PrivateKey,
+) (*Salesman, error) {
```
sshaman1101: this is awful, maybe will be better to pass config structure or use options?

insonmnia/miner/salesman.go
```diff
@@ -0,0 +1,343 @@
...
+		if plan.DealID.String() == dealID {
+			return plan, nil
+		}
+	}
+	return nil, fmt.Errorf("ask plan for deal id %s is not found", dealID)
+}
+
+func (m *Salesman) syncRoutine(ctx context.Context) {
+	m.log.Debugf("starting sync routine")
+	ticker := util.NewImmediateTicker(time.Second)
```
sshaman1101: `defer ticker.Stop()`.

insonmnia/miner/server.go
```diff
@@ -123,7 +126,16 @@ func NewMiner(cfg *Config, opts ...Option) (m *Miner, err error) {
 	}
 
 	if o.dwh == nil {
-		o.dwh = dwh.NewDumbDWH(o.ctx)
+		_, TLSConfig, err := util.NewHitlessCertRotator(context.Background(), o.key)
```
sshaman1101: We already have initialized gRPC credentials into the `cmd/worker/main.go`. To prevent double work you can pass it via options.

insonmnia/resource/pool.go
```diff
@@ -1,116 +1,186 @@
...
+func (m *Scheduler) ReleaseTask(taskID string) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	askPlanID, ok := m.taskToAskPlan[taskID]
+	if !ok {
+		return fmt.Errorf("could not find corresponding ask plan id for task %s", taskID)
+	}
+	pool, ok := m.askPlanPools[askPlanID]
+	if !ok {
+		return fmt.Errorf("could not consume resources for task - ask Plan with id %s not found", askPlanID)
```
sshaman1101: s/consume/release ?

sshaman1101:
As discussed privately, here is bit more things to fix:
* now all of the startup routines are disabled\commented, need to invent some method to perform startup synchronization, refreshes deals and orders status and check that matcher operates on actual data.
* Resources cleanup into `Scheduler.ReleaseTask` and `pool.release`.

insonmnia/benchmarks/benchmarks.go
```diff
@@ -87,15 +89,15 @@ func (bl *benchmarkList) readResults(ctx context.Context, reader io.ReadCloser)
 		return fmt.Errorf("cannot decode JSON response: %v", err)
 	}
 
-	var max uint64
+	bl.byID = make([]*pb.Benchmark, len(data))
```
3Hren: This logic is wrong when we remove some benchmarks or enumerate them using random offset.

insonmnia/hardware/hardware.go
```diff
@@ -59,6 +60,17 @@ func (h *Hardware) Hash() string {
 	return h.devicesMap().Hash()
 }
 
+func (m *Hardware) HashGPU(indexes []uint64) ([]string, error) {
+	hashes := []string{}
```
3Hren: Can preallocate.

insonmnia/miner/salesman.go
```diff
@@ -0,0 +1,343 @@
...
+		m.log.Warnf("could not save ask plan %s in storage - %s", plan.ID, err)
+		// TODO: log, what else can we do?
+		return
+	}
+	go m.waitForDeal(ctx, ordOrErr.Order)
+}
+
+func (m *Salesman) waitForDeal(ctx context.Context, order *sonm.Order) {
+	// TODO: make configurable
+	ticker := util.NewImmediateTicker(time.Second * 10)
```
3Hren: `defer ticker.Stop()`

insonmnia/miner/server.go
```diff
@@ -714,6 +729,9 @@ func (m *Miner) RunBenchmarks() error {
 		return err
 	}
 
+	log.S(context.Background()).Errorf("WTF??????")
+	PIDOR, _ := yaml.Marshal(m.hardware)
```
3Hren: You.

insonmnia/hardware/hardware.go
antmat: ok

insonmnia/miner/salesman.go
```diff
@@ -0,0 +1,343 @@
...
+		dealChan:  make(chan *sonm.Deal, 1),
+	}
+
+	if err := s.restoreState(); err != nil {
+		return nil, err
+	}
+	return s, nil
+}
+
+func (m *Salesman) Run(ctx context.Context) chan *sonm.Deal {
```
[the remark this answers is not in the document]
antmat: ok

insonmnia/miner/server.go
```diff
@@ -409,13 +449,18 @@ func (m *Miner) Start(ctx context.Context, request *pb.MinerStartRequest) (*pb.M
...
 
-	cGroup, resourceHandle, err := m.consume(request.GetOrderId(), resources)
-	if err != nil {
-		return nil, status.Errorf(codes.ResourceExhausted, "failed to start %v", err)
+	//TODO: generate ID
+	if err := m.resources.ConsumeTask(ask.ID, request.Id, request.Resources); err != nil {
+		return nil, fmt.Errorf("could not start task - %s", err)
 	}
+	defer func() {
+		if err != nil {
```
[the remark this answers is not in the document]
antmat: this is not the case here, but ok, this is doubtfull technique

insonmnia/hardware/hardware.go
```diff
@@ -68,6 +80,94 @@ func (h *Hardware) SetNetworkIncoming(IPs []string) {
...
+	result.Network.Incoming = h.Network.Incoming
+	//TODO: Make network device use DataSizeRate
+	result.Network.ThroughputIn.BitsPerSecond = h.Network.GetIn()
+	result.Network.ThroughputOut.BitsPerSecond = h.Network.GetOut()
+	return result
+}
+
+func insertBench(to []uint64, bench *sonm.Benchmark, proportion float64) ([]uint64, error) {
+	for len(to) <= int(bench.ID) {
+		to = append(to, uint64(0))
```
noxiouz: Not sure, but may be this is more clear:
```go
to = append(to, make([]uint64, len(to) - int(bench.ID))...)
```

insonmnia/miner/server.go
```diff
@@ -209,43 +225,68 @@ func NewMiner(cfg *Config, opts ...Option) (m *Miner, err error) {
...
+	//TODO: this is racy, because of post initialization of hardware via benchmarks
+	m.resources = resource.NewScheduler(o.ctx, m.hardware)
+
+	salesman, err := NewSalesman(o.ctx, o.storage, m.resources, m.hardware, o.eth, matcher, o.key)
+	if err != nil {
+		return nil, err
+	}
+	m.salesman = salesman
+
+	m.salesman.Run(o.ctx)
```
noxiouz: Run() returns the channel. Should we use it somehow?

insonmnia/miner/salesman.go
```diff
@@ -0,0 +1,343 @@
...
+func (m *Salesman) dropCGroup(planID string) error {
+	cgroup, ok := m.askPlanCGroups[planID]
+	if !ok {
+		return fmt.Errorf("cgroup for ask plan %s not found, probably no such plan", planID)
+	}
+	err := cgroup.Delete()
+	if err != nil {
+		return fmt.Errorf("could not drop cgroup %s for ask plan %s - %s", cgroup.Suffix(), planID, err)
+	}
+	delete(m.askPlanCGroups, planID)
```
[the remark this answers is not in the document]
3Hren: For example when it's already unmounted, or when suddenly your app lose its capabilities. Well, these cases should be really rare, but in almost any case we can't do anything without external access.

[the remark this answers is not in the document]
antmat: so, what's the hazard of leaving it in that map of cgroups? I'm just curious

insonmnia/hardware/hardware.go
antmat: you have a bug in your example)

insonmnia/miner/server.go
noxiouz: My bad

insonmnia/benchmarks/benchmarks.go
3Hren: Emm, what?

For example you have a json of:
```json
{
  "b1": {"id": 0},
  "b2": {"id": 2},
}
```

It's not malformed. Why it is? Otherwise we should reenumarate all benchmarks just because one day we removed "id": 1, in the middle, because of reasons.
Moreover, all orders that are currently in the market should be invalidated.
This is shit.

insonmnia/hardware/hardware.go
noxiouz: My bad

antmat:
The logic is unchenged here
See https://github.com/sonm-io/core/blob/master/insonmnia/benchmarks/benchmarks.go#L97

insonmnia/hardware/hardware.go
```diff
@@ -68,6 +80,94 @@ func (h *Hardware) SetNetworkIncoming(IPs []string) {
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			return nil, fmt.Errorf("unknown hash in passed resources - %s", hash)
+		}
+	}
+
+	hwBenches = append(hwBenches,
```
3Hren: I don't like this. Looks very error prone.

antmat:
It's the subject of our assumptions, because
```
{
  "ti" : "pidor"
}
```
is also valid json, but is not suitable for benchmark list

insonmnia/miner/config.go
```diff
@@ -41,6 +41,7 @@ type Config struct {
 	Benchmarks        benchmarks.Config   `yaml:"benchmarks"`
 	Whitelist         WhitelistConfig     `yaml:"whitelist"`
 	MetricsListenAddr string              `yaml:"metrics_listen_addr" default:"127.0.0.1:14000"`
+	DWHEndpoint       string              `yaml:"dwh" default:"5.178.85.52:15021"`
```
3Hren: Leaked endpoint.

insonmnia/miner/salesman.go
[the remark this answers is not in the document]
3Hren: It will be there forever.

insonmnia/miner/salesman.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+}
+
+func (m *Salesman) syncWithBlockchain(ctx context.Context) {
+	m.log.Debugf("syncing salesman with blockchain")
+	plans := m.state.AskPlans()
+	for _, plan := range plans {
+		orderId := plan.GetOrderID()
+		dealId := plan.GetDealID()
+		if !dealId.IsZero() {
```
3Hren: Double checking for both IDs. Smells.

insonmnia/miner/salesman.go
```diff
@@ -0,0 +1,345 @@
...
+		m.log.Warnf("could not get benchmarks for ask plan %s - %s", plan.ID, err)
+		return
+	}
+	net := m.hardware.Network
+	order := &sonm.Order{
+		OrderType:      sonm.OrderType_ASK,
+		OrderStatus:    sonm.OrderStatus_ORDER_ACTIVE,
+		AuthorID:       sonm.NewEthAddress(crypto.PubkeyToAddress(m.ethkey.PublicKey)),
+		CounterpartyID: plan.GetCounterparty(),
+		Duration:       uint64(plan.GetDuration().Unwrap().Seconds()),
```
3Hren: As I remember we put here nanoseconds, aren't we?

insonmnia/miner/salesman.go
```diff
@@ -0,0 +1,345 @@
...
+		Price:          plan.GetPrice().GetPerSecond(),
+		//TODO:refactor NetFlags in separqate PR
+		Netflags:      sonm.NetflagsToUint([3]bool{net.GetOverlay(), true, net.Incoming}),
```
3Hren: `net.GetOverlay()` returns bool?

insonmnia/miner/salesman.go
```diff
@@ -0,0 +1,345 @@
...
+
+		if err := m.state.SaveAskPlan(plan); err != nil {
+			m.log.Warnf("could not get save ask plan with deal %s - %s", order.DealID.Unwrap().String(), err)
+			// TODO: log, what else can we do?
+			return
+		}
+	}
+}
+
+func (m *Salesman) placeOrder(ctx context.Context, plan *sonm.AskPlan) {
```
3Hren: Why can't we return error here?

insonmnia/resource/pool.go
```diff
@@ -1,116 +1,187 @@
...
-		usage.NumGPUs = len(p.OS.GPU)
+func (p *pool) getFree() (*sonm.AskPlanResources, error) {
+	res := deepcopy.Copy(p.all).(*sonm.AskPlanResources)
+	usage, err := p.getUsage()
+	if err != nil {
+		return nil, err
 	}
+	err = res.Sub(usage)
+	if err != nil {
+		pool, _ := yaml.Marshal(res)
```
3Hren: Why to ignore errors?

insonmnia/resource/pool.go
```diff
@@ -1,116 +1,187 @@
...
 }
 
-func (p *Pool) Release(usage *Resources) {
-	p.mu.Lock()
-	defer p.mu.Unlock()
-
-	p.usage.NumCPUs -= usage.NumCPUs
-	p.usage.Memory -= usage.Memory
-	p.usage.NumGPUs -= usage.NumGPUs
+func (p *pool) release(ID string) error {
```
3Hren: Just `delete(p.used, ID)`?

proto/ask_plan.go
```diff
@@ -37,9 +46,224 @@ func (m *AskPlan) Validate() error {
...
+	m.Storage.Size.Bytes -= resources.GetStorage().GetSize().GetBytes()
+	m.GPU.Sub(resources.GetGPU())
+	m.Network.ThroughputIn.BitsPerSecond -= resources.GetNetwork().GetThroughputIn().GetBitsPerSecond()
+	m.Network.ThroughputOut.BitsPerSecond -= resources.GetNetwork().GetThroughputOut().GetBitsPerSecond()
+	return nil
+}
+
+func (m *AskPlanResources) ToHostConfigResources(cgroupParent string) container.Resources {
+	//TODO: check and discuss
+	return container.Resources{
```
3Hren: What the fuck is this? Check the code you've removed.

proto/hub.proto
```diff
@@ -108,3 +108,44 @@ message DealInfoReply {
...
+    Status status = 1;
+    string imageName = 2;
+    string ports = 3;
+    uint64 uptime = 4;
+    ResourceUsage usage = 5;
+    AskPlanResources allocatedResources = 6;
+    string minerID = 7;
+}
+
+//message AvailableResources {
```
3Hren: ?

proto/hub.proto
```diff
@@ -108,3 +108,44 @@ message DealInfoReply {
...
+//    int64 MemoryReservation = 12;
+//    int64 MemorySwap = 13;
+//    int64 PidsLimit = 14;
+//    string cgroup = 15;
+//    string cgroupParent = 16;
+//}
+
+message StatusMapReply {
+    map<string, TaskStatusReply> statuses = 1;
+}
```
3Hren: Newline.

vendor/vendor.json
```diff
@@ -1208,6 +1208,8 @@
 			"path": "github.com/mohae/deepcopy",
 			"revision": "c48cc78d482608239f6c4c92a4abd87eb8761c90",
 			"revisionTime": "2017-09-29T03:49:55Z"
+		},
+		{
```
3Hren: Put in the separate PR.

insonmnia/hardware/hardware.go
antmat: any better ideas?)

insonmnia/hardware/hardware.go
antmat: alternative is tons of err!=nil checking

insonmnia/hardware/hardware.go
```diff
@@ -68,6 +80,94 @@ func (h *Hardware) SetNetworkIncoming(IPs []string) {
...
+		if bench.Result < to[bench.ID] {
+			to[bench.ID] = bench.Result
+		}
+	}
+	return to, nil
+}
+
+func (h *Hardware) ResourcesToBenchmarks(resources *sonm.AskPlanResources) (*sonm.Benchmarks, error) {
+	if !resources.GPU.Normalized() {
+		return nil, errors.New("passed resources are not normalized, call resources.GPU.Normalize(hardware) first")
```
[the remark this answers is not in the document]
antmat: I've thought about this, but I don't like mutating passed argument

insonmnia/miner/salesman.go
antmat: no we didn't, we use nanoseconds for our Duration type, however bc market steal work with serconds

insonmnia/miner/salesman.go
antmat: yep, I will use finally available GetOutbound and GetIncoming() for consistency

insonmnia/miner/salesman.go
antmat: it's a private function and we can not do anything with this error, but ok, let's make it more idiomatic

insonmnia/resource/pool.go
antmat: because it is marshalling for error printing (and this should never fail)

insonmnia/resource/pool.go
antmat: I think here it's a good idea to signal about pool inconsistency. If you don't care - you can ignore this error

proto/ask_plan.go
antmat: fixed

vendor/vendor.json
antmat: done

quasisamurai:
:hankey:

3Hren:
Also fix tasks authorization.

proto/ask_plan.go
3Hren: Where?

proto/ask_plan.go
```diff
@@ -37,9 +46,221 @@ func (m *AskPlan) Validate() error {
...
+	return m.GetResources().GetGPU().Validate()
+}
+
+func NewEmptyAskPlanResources() *AskPlanResources {
+	res := &AskPlanResources{}
+	res.InitNilWithZero()
+	return res
+}
+
+func (m *AskPlanResources) InitNilWithZero() {
```
3Hren: Make private.

insonmnia/resource/pool.go
3Hren: But it can fail.

proto/ask_plan.go
antmat: I mean I've put exactly the same code as it was there previously

insonmnia/resource/pool.go
antmat: I really don't get it, what is your suggestion? This is used only to format error and in worst case we will have badly formatted error and that's all. btw it can fail only if some non-marshalable structure is passed to yaml.Marshal, which is not our case - overvise that would indicate logic error. I can put panic there of course if you want

insonmnia/miner/salesman.go
oopcode: i think that a config structure is better in terms of readability, but nevermind

insonmnia/cgroups/cgroup_linux.go
```diff
@@ -83,19 +81,23 @@ func initializeControlGroup(name string, resources *specs.LinuxResources) (CGrou
...
 	if err = control.Update(resources); err != nil {
-		return nil, fmt.Errorf("failed to set resource limit on parent cgroup %s: %v", name, err)
+		return nil, fmt.Errorf("failed to set resource limit on cgroup %s: %v", name, err)
 	}
 
 	return &cgroup{control, name}, nil
 }
+
+func initializeControlGroup(name string, resources *specs.LinuxResources) (CGroup, error) {
+	return newCGroup("/"+name, resources)
```
sshaman1101: Is there any reason not to do this into the `newCGroup`  function? Why we need an extra function just for appending a slash?

PR #674, opened 2018-04-29
refact: dwh uses pool for slices of filters

noxiouz:
Reuse allocated slices of filters. It's rather heavy work for GC as *filter is pointer and contains Value interface.


3Hren:
I have no idea whether this improves something. Maybe prove with benchmarks results?

PR #675, opened 2018-04-29
fix: dwh sql bugs and sqlite-specific commands

oopcode:
* all references to `rowid` are removed,
* fixed missing `Unwrap()` for EthAddress


insonmnia/dwh/server.go
```diff
@@ -998,6 +999,7 @@ func (w *DWH) onDealOpened(dealID *big.Int) error {
 
 	_, err = tx.Exec(
 		w.commands["insertDealCondition"],
+		nil,
```
3Hren: Calling methods like `doSomething(nil, nil, nil, nil, ...)` clearly don't improve readability, so I for example have no idea what's going on here.

oopcode: this will go away with the Postgres PR, it was an intermediate step.

PR #678, opened 2018-05-01
feat: integrate Matcher into Node

sshaman1101:


etc/node.yaml
```diff
@@ -26,6 +26,12 @@ dwh:
   # marketplace gRPC endpoint, required
   endpoint: "8125721C2413d99a33E351e1F6Bb4e56b6b633FD@127.0.0.1:15021"
 
+# fine tunes for Matcher
+# matcher:
+#  disabled: false
```
3Hren: I'd rather not unlikely not to do so.

Anyway, comment the whole section to disable matching.

PR #687, opened 2018-05-03
refactor: dead code elimination

3Hren:


3Hren:
Oops, should be in another branch.

PR #689, opened 2018-05-03
refactor: proper marshalling for EthAddress

sshaman1101:
This PR implements `MarshalText` and `UnmarshalText` methods for the `EthAddress` type.
It allows satisfying both JSON and YAML marshallers requirements.

### DIFF
--- a/proto/insonmnia.go
+++ b/proto/insonmnia.go
@@ -1,7 +1,6 @@
 package sonm
 
 import (
-	"errors"
 	"fmt"
 	"math/big"
 	"strings"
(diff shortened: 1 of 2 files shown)


sshaman1101:
@oopcode PTAL, seems that (un)marshaling is used somewhere [in the DWH](https://github.com/sonm-io/core/blob/master/insonmnia/dwh/server_test.go#L741).

proto/insonmnia.go
```diff
@@ -54,25 +53,19 @@ func (m *EthAddress) Unwrap() common.Address {
 	return common.BytesToAddress(m.Address)
 }
 
-func (m *EthAddress) MarshalYAML() (interface{}, error) {
-	if m != nil {
-		return m.Unwrap().Hex(), nil
-	}
-	return nil, nil
+func (m *EthAddress) MarshalText() ([]byte, error) {
```
3Hren: Seems like we can also pass by value here.

proto/insonmnia.go
```diff
@@ -54,25 +53,19 @@ func (m *EthAddress) Unwrap() common.Address {
...
-	var v string
-	if err := unmarshal(&v); err != nil {
-		return err
-	}
-
+func (m *EthAddress) UnmarshalText(text []byte) error {
+	v := string(text)
 	if !common.IsHexAddress(v) {
-		return errors.New("invalid ethereum address format")
+		return fmt.Errorf("invalid ethereum address format \"%s\"", v)
```
3Hren: What do you expect here to see? What if it's not a string, but unprintable byte array?

sshaman1101: It used just for notifications of human (when parsing YAMLs). 
If an error message will include the wrong value it will be easier to detect where's it comes from.

PR #704, opened 2018-05-07
refactor: move blockchain-related code from CLI to Node

sshaman1101:


proto/node.proto
```diff
@@ -81,3 +88,16 @@ message Worker {
...
+    // into live-chian ethereum network.
+    rpc TestTokens(Empty) returns (Empty) {}
+    // Balance provide account balance for live- and side- chains.
+    rpc Balance(Empty) returns (BalanceReply) {}
+}
+
+message BalanceReply {
+    BigInt liveBalance = 1;
+    BigInt sideBalance = 2;
+}
```
3Hren: Newline.

sshaman1101: done!

PR #722, opened 2018-05-10
fix: event re-processing & better logging

oopcode:
This PR fixes re-processing of events from the last block by making blockchain API send logs only if new blocks are present.


blockchain/api.go
```diff
@@ -696,22 +696,27 @@ func (api *BasicEventsAPI) GetEvents(ctx context.Context, fromBlockInitial *big.
 					api.logger.Info("no logs, skipping")
 					continue
 				}
-				fromBlock = big.NewInt(int64(logs[numLogs-1].BlockNumber))
-
-				var eventTS uint64
-				for _, log := range logs {
-					logBlockNumber := big.NewInt(int64(log.BlockNumber))
-					if lastBlockNumber.Cmp(logBlockNumber) != 0 {
```
antmat: Looks like the problem was in "==" comparision instead of ">=". I don't get the idea of adding some strange double check.

oopcode: there was no check in the first place, events were emitted from logs unconditionally

PR #723, opened 2018-05-10
fix: update deal total payout (DWH)

oopcode:
Old version only updated payout for deal conditions, this PR adds updates for corresponding deals.
(P.S. And missed mutex Lock/Unlock).

### DIFF
--- a/insonmnia/dwh/server.go
+++ b/insonmnia/dwh/server.go
@@ -1328,6 +1328,15 @@ func (w *DWH) onDealChangeRequestUpdated(eventTS uint64, changeRequestID *big.In
 }
 
 func (w *DWH) onBilled(eventTS uint64, dealID, payedAmount *big.Int) error {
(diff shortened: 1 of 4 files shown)


insonmnia/dwh/server.go
```diff
@@ -1328,6 +1328,9 @@ func (w *DWH) onDealChangeRequestUpdated(eventTS uint64, changeRequestID *big.In
 }
 
 func (w *DWH) onBilled(eventTS uint64, dealID, payedAmount *big.Int) error {
+	w.mu.Lock()
```
antmat: This function doesn't fit in two screens(which is a problem itself), and is being processed under lock. BTW I don't get why do you even need it.

oopcode: what does function size have to do with this PR? processing this event requires exactly N things to happen, and those things do not occur anywhere else in the code; I can’t see any way to reduce this code, if you have any ideas — please inform me

PR #725, opened 2018-05-10
fix: getProfile by Name works as SQL 'LIKE' (DWH)

oopcode:


antmat:
Any motivation for doing this?

oopcode:
it’s a request from ui team

PR #741, opened 2018-05-13
fix: return a corresponding error for non-existent orders and deals

sshaman1101:


sshaman1101:
Unfortunately, the blockchain API does not return the ID of entity, so I've forced to check for some required fields. 💩

PR #754, opened 2018-05-14
fix: 'worker status' count only running tasks

sshaman1101:


insonmnia/miner/server.go
```diff
@@ -604,13 +604,22 @@ func (m *Miner) Stop(ctx context.Context, request *pb.ID) (*pb.Empty, error) {
 	return &pb.Empty{}, nil
 }
 
-func (m *Miner) CollectTasksStatuses() map[string]*pb.TaskStatusReply {
+func (m *Miner) CollectTasksStatuses(statuses ...pb.TaskStatusReply_Status) map[string]*pb.TaskStatusReply {
```
3Hren: What will be if a pass `TaskStatusReply_RUNNING` for example twice?

3Hren: Okay, seems like nothing bad.

PR #755, opened 2018-05-14
fix: use gostyle optional for benchmarks

antmat:
I don't know how to emulate optionals in a better way, but they are needed for min summarizing algo.
@sshaman1101 @oopcode @3Hren PTAL


3Hren:
Not looking at the code itself, using pointer to integer for emulating optionals is completely crap. Please investigate some other way to do this or you will be ashamed for the entire eternity and be called as eternal pidor.

PR #775, opened 2018-05-18
feat: master-worker relationships

antmat:
This PR changes master-worker relations stubs to actual code. 
Master file din config is introduced, which if set forces worker to send master request to blockchain and wait for approval.
@oopcode @sshaman1101 @3Hren PTAL


cmd/cli/commands/master.go
```diff
@@ -21,19 +24,27 @@ var masterRootCmd = &cobra.Command{
...
+	Use:    "list [master_eth]",
+	Short:  "Show known worker's addresses",
+	PreRun: loadKeyStoreIfRequired,
+	Run: func(cmd *cobra.Command, args []string) {
 		ctx, cancel := newTimeoutContext()
 		defer cancel()
 
+		var master common.Address
+		if len(args) > 0 {
+			master = common.HexToAddress(args[0])
```
sshaman1101: we should validate that given value is valid address using `common.IsHexAddress(s)`

PR #777, opened 2018-05-18
feat: stale events processing (DWH)

oopcode:
* L1 DWH doesn't keep stale entities (order, deals, etc.). This PR introduces code that avoids fully processing events that point to stale entities (by saving stale ids to a separate table).
* Suppressing errors related to uniqueness constraints was removed.


insonmnia/dwh/server.go
```diff
@@ -1157,3 +1217,40 @@ func (w *DWH) checkBenchmarks(benches *pb.Benchmarks) error {
 
 	return nil
 }
+
+func (w *DWH) removeEntityID(id *big.Int, entity string) error {
```
sshaman1101: The name is weird, maybe will be better to use smth like `removeSlateEntityByID`?

insonmnia/dwh/storage.go
```diff
@@ -56,6 +56,9 @@ type storage interface {
 	GetLastKnownBlock(conn queryConn) (uint64, error)
 	InsertLastKnownBlock(conn queryConn, blockNumber int64) error
 	UpdateLastKnownBlock(conn queryConn, blockNumber int64) error
+	StoreID(conn queryConn, id *big.Int, entity string) error
+	RemoveID(conn queryConn, id *big.Int, entity string) error
+	CheckID(conn queryConn, id *big.Int, entity string) (bool, error)
```
sshaman1101: These three methods names are not really obvious, names explain actions, but not an entity to act with. 
I'd personally prefer to make it bit more self-describing (like all previous methods do).

PR #778, opened 2018-05-21
 feat: transaction confirmation awaiting

Sokel:
* add utils that await block confirmation

* upd PlaceOrder & OpenDeal funcs

* add custom ethclient

### DIFF
--- a/blockchain/api.go
+++ b/blockchain/api.go
@@ -12,7 +12,6 @@ import (
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/noxiouz/zapctx/ctxlog"
 	"github.com/pkg/errors"
 	"github.com/sonm-io/core/blockchain/market"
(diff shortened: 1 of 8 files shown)


blockchain/api.go
```diff
@@ -141,7 +140,7 @@ func NewAPI(opts ...Option) (API, error) {
 		return nil, err
 	}
 
-	marketApi, err := NewBasicMarket(clientSidechain, market.MarketAddr(), defaults.gasPriceSidechain, defaults.logParsePeriod)
+	marketApi, err := NewBasicMarket(clientSidechain, market.MarketAddr(), defaults.gasPriceSidechain, defaults.logParsePeriod, 4)
```
antmat: make it const at least

blockchain/gen_receipt_json.go
```diff
@@ -0,0 +1,93 @@
+// Code generated by github.com/fjl/gencodec. DO NOT EDIT.
```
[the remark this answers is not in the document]
Sokel: Its advanced json marshaller with features which the standard json package cannot offer.
Any docs here:
https://godoc.org/github.com/fjl/gencodec

blockchain/receipt.go
```diff
@@ -0,0 +1,218 @@
+// Copyright 2014 The go-ethereum Authors
```
antmat: what is the difference with the original code?

blockchain/util.go
```diff
@@ -80,22 +78,61 @@ func findLogByTopic(ctx context.Context, client *ethclient.Client, tx *types.Tra
...
 
-			return log, err
+			return logs, err
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
+	}
+}
+
+func WaitTransactionReceipt(ctx context.Context, client CustomEthereumClient, confirmations int64, logParsePeriod time.Duration, tx *types.Transaction) (*Receipt, error) {
```
antmat: it's rather unclear from the name of this method - what is the difference with the waitForTransactionResult

blockchain/gen_receipt_json.go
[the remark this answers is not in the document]
antmat: Is it really worth introducing custom generated non-obvious code from nsome strange tool just for checking several fields for nil?

blockchain/receipt.go
Sokel: only on `BlockNumber` field

antmat: How about smth like 
```
type SokelReciept struct {
 eth.Receipt
 BlockNumber uint64
}
```

Sokel: Thanks a lot!

blockchain/util.go
Sokel: `WaitTransactionReceipt` use Custom client and waiting correct Receipt after confirmations
In other, `waitTransactionResult` stay this as deprecated stuff.
Thanks i would plan to remove this.

Sokel: diff so strange, code changed only 3 lines

Sokel:
@sshaman1101 @antmat PTAL

PR #792, opened 2018-05-22
refact: new worker management approach

antmat:
Changed:
  * worker keys are now generated and stored in boltDB
  * node can now connect to any worker specified in context
  * cli takes current worker address from config

 TODO: change current worker addr with switching command


insonmnia/node/tasks.go
```diff
@@ -49,28 +49,6 @@ func (t *tasksAPI) List(ctx context.Context, req *pb.TaskListRequest) (*pb.TaskL
 	return reply, nil
 }
 
-func (t *tasksAPI) getSupplierTasks(ctx context.Context, tasks map[string]*pb.TaskStatusReply, deal *pb.Deal) {
```
3Hren: Out of scope.

util/grpc.go
```diff
@@ -4,6 +4,10 @@ import (
 	"strings"
 )
 
+const (
+	WorkerAddressHeader = "x_worker_eth_addr"
```
3Hren: Rename to `x-worker-eth-addr`, because of [convention](https://http2.github.io/http2-spec/compression.html#static.table.definition).

insonmnia/node/tasks.go
antmat: Nope, I've succesfully refactored getWorkerClientByEthAddr usage

PR #817, opened 2018-05-24
feat: add SimpleGatekeeper contract drop Gatekeeper contract

Sokel:
Add SimpleGate - its used in testnet release.

Drop Gatekeeper contract - api changed.


blockchain/source/migrations/4_init_gatekeeper.js
```diff
@@ -0,0 +1,23 @@
...
+    if (network === 'private') {
+        SNM.deployed()
+            .then(function (res) {
+                res.transfer(SimpleGatekeeper.address, 444 * 1e6 * 1e18, { gasPrice: 0 });
+            })
+            .catch(function (err) {
+                console.log(err);
+            });
+    } else if (network === 'rinkeby') {
+    } else {
```
sshaman1101: wow, such `else if`, so empty

blockchain/source/contracts/SimpleGatekeeper.sol
```diff
@@ -0,0 +1,47 @@
...
+    constructor(address _token) public {
+        token = StandardToken(_token);
+        owner = msg.sender;
+    }
+
+    uint256 public transactionAmount = 0;
+    mapping(bytes32 => bool) public paid;
+
+    event PayInTx(address indexed from, uint256 indexed txNumber, uint256 indexed value);
+    event PayoutTx(address indexed from, uint256 indexed txNumber, uint256 indexed value);
```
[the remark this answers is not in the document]
Sokel: I don't think so.

blockchain/source/migrations/4_init_gatekeeper.js
Sokel: it's ok. Deployer must support this section.

Sokel:
> What is the difference between gatekeeper and simpleGatekeeper?
they have one functional but different implementations and relevant to different releases

PR #833, opened 2018-05-25
refactor: start task is splitted to settings and dealID

antmat:
* start task request is split in order to load and validate only settings from file
  * Deal field is collapsed to DealID, as only id is usefull
  * structs.TaskStartRequest is dropped in favour of Validate method
@3Hren @oopcode @sshaman1101 PTAL


proto/worker.proto
```diff
@@ -51,14 +52,18 @@ service Worker {
...
+    AskPlanResources resources = 3;
+}
+
 message StartTaskRequest {
     // Deal points to the deal associated with workers where the task should be
     // started.
-    Deal deal = 1;
+    BigInt dealID = 1;
+    TaskSpec spec = 2;
     // Container describes container settings.
```
3Hren: Comment?

PR #838, opened 2018-05-25
chore: installing node modules first

Sokel:
Move `node_modules` to installing node module before build.


Sokel:
travis likes to drop the installing dependencies via `apt-get` or `npm`.
this will save us some time if this happens before the build.

PR #839, opened 2018-05-27
refactor: less boilerplate for blockchain options

sshaman1101:
This PR extracts geth-node connect options into re-usable structure.
Externally managed options now represented as two similar sets
of chainOpts - for side and live chains.

Subsets of the Blockchain's API constructors (like TokenAPI or MarketAPI) now accepts
only token address and generalized chainOpts structure. HTTP client to communicate with
geth-node now constructing by chainOpts itself, no need to explicitly pass it to API instance.

`initEthClient` and `initCustomEthClient` are dropped. `CustomEthereumClient` used
everywhere because customEthClient have embedded ethClient. We can easily use only one.

Also, naming eventually fixed: `sidechain` and `livechain` words added to some
funcs, consts, and structs to explicitly say to what they are related for.


blockchain/options.go
```diff
@@ -3,70 +3,94 @@ package blockchain
...
+}
+
 type options struct {
-	gasPrice             int64
-	gasPriceSidechain    int64
-	apiEndpoint          string
-	apiSidechainEndpoint string
-	logParsePeriod       time.Duration
-	blockConfirmations   int64
+	livechain *chainOpts
```
Sokel: `livechain` is abuse naming.
we have agreement to name this `masterchain`.

Sokel:
> initEthClient and initCustomEthClient are dropped. CustomEthereumClient used
> everywhere because customEthClient have embedded ethClient. We can easily use only one.

Any methods in `CustomEthereumClient` might be changed and planned to changed. 
You sure to do this?

blockchain/options.go
```diff
@@ -3,70 +3,94 @@ package blockchain
...
 
 func defaultOptions() *options {
 	return &options{
-		gasPrice:             defaultGasPrice,
-		gasPriceSidechain:    defaultGasPriceSidechain,
-		apiEndpoint:          defaultEthEndpoint,
-		apiSidechainEndpoint: defaultSidechainEndpoint,
-		logParsePeriod:       time.Second,
-		blockConfirmations:   defaultBlockConfirmations,
+		livechain: &chainOpts{
```
Sokel: why options for masterchain doesn't contain `blockConfiramtions`?

blockchain/api.go
```diff
@@ -736,10 +732,15 @@ func (api *StandardTokenApi) TotalSupply(ctx context.Context) (*big.Int, error)
 type TestTokenApi struct {
 	client        EthereumClientBackend
 	tokenContract *marketAPI.SNMTToken
-	gasPrice      int64
+	opts          *chainOpts
 }
 
-func NewTestToken(client EthereumClientBackend, address common.Address, gasPrice int64) (TestTokenAPI, error) {
+func NewTestToken(address common.Address, opts *chainOpts) (TestTokenAPI, error) {
+	client, err := opts.newClient()
```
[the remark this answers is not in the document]
Sokel: every Client create new `http/Client` and dial with endpoint on init.
I think that this will set a lot of trouble now and in the future.

blockchain/options.go
```diff
@@ -3,70 +3,94 @@ package blockchain
...
-	defaultGasPrice           = 20000000000 // 20 Gwei
-	defaultGasPriceSidechain  = 0
+	defaultLivechainGasPrice  = 20000000000 // 20 Gwei
+	defaultSidechainGasPrice  = 0
 	defaultBlockConfirmations = 5
+	defaultLogParsePeriod     = time.Second
 )
 
+// chainOpts describes common options
+// for almost any geth-node connection
```
Sokel: the client does not connect to the geth node, he connects to the rpc endpoint

blockchain/options.go
```diff
@@ -3,70 +3,94 @@ package blockchain
...
+// (live Eth network, rinkeby, SONM sidechain
+// or local geth-node for testing).
+type chainOpts struct {
+	gasPrice           int64
+	endpoint           string
```
Sokel: its incorrect to includes params to connect to rpc and params to interact with blockchain in one struct

blockchain/options.go
sshaman1101: how many confirmations is applicable for masterChain settings?

blockchain/options.go
sshaman1101: I'll do it with upcoming PR to not make out-of-scope changes here.

blockchain/options.go
oopcode: why exactly

PR #845, opened 2018-05-27
fix: receipt unmarshalling

Sokel:
As far as ethereum Receipt has it's own unmarshaller blockNumber of our custom Receipt was never been unmarshalled. Introduced UnmarshallJSON method overrides this behaviour. //edited @antmat


blockchain/client.go
```diff
@@ -12,11 +13,35 @@ import (
...
+func (r *Receipt) UnmarshalJSON(input []byte) error {
+	// call parent unmarshal
+	err := r.Receipt.UnmarshalJSON(input)
+	if err != nil {
+		return err
+	}
+
+	// define pure structure for clearly casting
+	type Receipt struct {
+		BlockNumber string `json:"BlockNumber" gencodec:"required"`
```
antmat: any reason for those tags?

blockchain/client.go
```diff
@@ -12,11 +13,35 @@ import (
...
+// reference: go-ethereum/core/types/Receipt has custom Unmarshaler for casting for its own types
+func (r *Receipt) UnmarshalJSON(input []byte) error {
+	// call parent unmarshal
+	err := r.Receipt.UnmarshalJSON(input)
+	if err != nil {
+		return err
+	}
+
+	// define pure structure for clearly casting
+	type Receipt struct {
```
antmat: name shadowing

blockchain/examples/customClient/customClient.go
```diff
@@ -40,7 +40,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	rec, err := blockchain.WaitTransactionReceipt(context.TODO(), client, time.Duration(3*time.Second), tx)
+	rec, err := blockchain.WaitTransactionReceipt(context.TODO(), client, 4, time.Duration(3*time.Second), tx)
```
antmat: out of the scope

blockchain/client.go
```diff
@@ -12,11 +13,35 @@ import (
...
+		BlockNumber string `json:"BlockNumber" gencodec:"required"`
+	}
+	var dec Receipt
+	if err := json.Unmarshal(input, &dec); err != nil {
+		return err
+	}
+
```
antmat: Please add check for non-empty block number as far as it is required

blockchain/client.go
```diff
@@ -12,11 +14,39 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
+// Receipt extends go-ethereum/core/types/Receipt struct with BlockNumber
 type Receipt struct {
 	*types.Receipt
 	BlockNumber string
```
sshaman1101: Looks like a field named `BlockNumber` must have a numeric type. 
This field is using to compare with block confirmation, and both is type-casted into `big.Int`, that is totally meaningless. I'd prefer to represent it as uint64.

blockchain/client.go
```diff
@@ -12,11 +14,39 @@ import (
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
+// Receipt extends go-ethereum/core/types/Receipt struct with BlockNumber
```
sshaman1101: I think that path to a type is completely useless.

blockchain/client.go
```diff
@@ -12,11 +14,39 @@ import (
...
 type Receipt struct {
 	*types.Receipt
 	BlockNumber string
 }
 
+// UnmarshalJSON implement json/Unmarshaler interface (stdlib)
+// reference: go-ethereum/core/types/Receipt has custom Unmarshaler for casting for its own types
```
sshaman1101: Do not specify **what** happen into a method, specify **why do we need to do this**.
I personally feel that this comment is weird and meaningless.

blockchain/client.go
```diff
@@ -12,11 +14,40 @@ import (
...
+	err := r.Receipt.UnmarshalJSON(input)
+	if err != nil {
+		return err
+	}
+
+	// define pure structure for clearly casting
+	type rec struct {
+		BlockNumber string `json:"BlockNumber"`
+	}
+	var dec rec
```
antmat: (Imagine "Wilhelm Tell" theme)
dec rec, dec rec, dec rec, dec rec
dec rec, dec rec, dec rec, dec rec
dec rec, dec rec, dec rec, dec rec
dec reeeeeec, dec rec, dec rec

PR #862, opened 2018-05-28
feat: moving profiles to core

quasisamurai:
Moved ProfilesRegistry to core.


sportloto:
@Queen93rinnies You dev too?

quasisamurai:
@sportloto 
Ere we go, ere we go, ere we go.
Ere we go, ere we go,
cross the Kosmos.

Because you a da Orks, annit?

PR #888, opened 2018-05-30
chore: cleanup contract project after accidentally merge

Sokel:


blockchain/source/migrations/7_deploy_market.js
```diff
@@ -6,8 +6,13 @@ var ProfileRegistry = artifacts.require('./ProfileRegistry.sol');
 
 module.exports = function (deployer, network) {
     if (network === 'private') {
-        deployer.deploy(Market, '0x5db024c6469634f4b307135cb76e8e6806f007b3', Blacklist.address, OracleUSD.address,
-            '0xcc1cb65bdea124520dbdcc9e82b21b4352e45ad9', 12, { gasPrice: 0 });
+        deployer.deploy(Market,
+            '0x5db024c6469634f4b307135cb76e8e6806f007b3', // token address
+            Blacklist.address, // Blacklist address
+            OracleUSD.address, // Oracle address
+            '0xcc1cb65bdea124520dbdcc9e82b21b4352e45ad9', // ProfileRegistry address
```
sshaman1101: Why somewhere you use addresses from a variable, and somewhere address is hardcoded string?

quasisamurai: Cause we're not going to redeploy token and also profile registry

PR #929, opened 2018-06-01
chore: DWH startup refactoring

oopcode:
* Fixed an minor bug: DWH could crash with segfault after a `SIGINT` at the very startup. This was possible in "cold start" mode because http server was not setup yet and there was no checks for `nil` in the `Stop()` method.
* Made blockchain config required.
* Merged two redundant methods (`setupPostgres` and `setupSQLite`) into one method (plus some refactoring for DB setup).


cmd/dwh/main.go
```diff
@@ -52,7 +52,8 @@ func run() error {
 		w.Stop()
 	}()
 
-	log.G(ctx).Info("starting DWH service", zap.String("grpc_addr", cfg.GRPCListenAddr), zap.String("http_addr", cfg.HTTPListenAddr))
```
antmat: out of the scope. Besides imo one liner of such size is better.

insonmnia/dwh/server.go
```diff
@@ -137,42 +100,105 @@ func (m *DWH) Stop() {
...
+	}
+}
+
+func (m *DWH) setupDB() error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	numBenchmarks, err := m.blockchain.Market().GetNumBenchmarks(m.ctx)
+	if err != nil {
+		m.Stop()
```
3Hren: Deadlock here.

insonmnia/dwh/server.go
```diff
@@ -137,42 +100,105 @@ func (m *DWH) Stop() {
...
+		return errors.Wrap(err, "failed to GetNumBenchmarks")
+	}
+	if numBenchmarks >= NumMaxBenchmarks {
+		m.Stop()
```
3Hren: And here.

PR #930, opened 2018-06-01
feat: coldstart auto block number (DWH)

oopcode:
This PR makes use of `CustomEthereumClient`'s `GetLastBlock()` method to automatically determine when to build indices in cold start mode.


insonmnia/dwh/config.go
```diff
@@ -16,7 +16,7 @@ type Config struct {
 	Storage           *storageConfig     `yaml:"storage" required:"true"`
 	Blockchain        *blockchain.Config `yaml:"blockchain"`
 	MetricsListenAddr string             `yaml:"metrics_listen_addr" default:"127.0.0.1:14004"`
-	ColdStart         *ColdStartConfig   `yaml:"cold_start"`
+	ColdStart         bool               `yaml:"cold_start"`
```
antmat: do we really need this behavour to be configurable?

PR #935, opened 2018-06-03
feat/overseer: mark containers with deal ID

noxiouz:
Going to use to maintain limits for a deal (like storage space)


insonmnia/worker/overseer.go
```diff
@@ -28,6 +28,7 @@ import (
 )
 
 const overseerTag = "sonm.overseer"
+const dealIDaTag = "sonm.dealid"
```
3Hren: Typo?

PR #938, opened 2018-06-04
fix: use askplan hasher instead of hardware hasher 

antmat:
Now we use indexes from ask-plan to determine gpu hash instead of hardware. This allows to use GPU indexes as they appear in ask-plan rather then in hardaware

### DIFF
--- a/proto/ask_plan.go
+++ b/proto/ask_plan.go
@@ -213,6 +213,22 @@ type GPUHasher interface {
 	HashGPU(indexes []uint64) (hashes []string, err error)
 }
 
+type AskPlanHasher struct {
+	*AskPlanResources
+}
+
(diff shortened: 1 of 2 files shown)


proto/ask_plan.go
```diff
@@ -213,6 +213,22 @@ type GPUHasher interface {
...
 
+type AskPlanHasher struct {
+	*AskPlanResources
+}
+
+func (m *AskPlanHasher) HashGPU(indexes []uint64) ([]string, error) {
+	askPlanHashes := m.GetGPU().GetHashes()
+	resultHashes := make([]string, 0, len(indexes))
+	for _, idx := range indexes {
+		if idx >= uint64(len(askPlanHashes)) {
```
oopcode: computed each time, weird

PR #939, opened 2018-06-04
feat: use blacklists in GetMatchingOrders (DWH)

oopcode:
Filter orders that:
	1. have our Master/Author in their Master/Author/Blacklist blacklist,
	2. whose Master/Author is in our Master/Author/Blacklist blacklist.


3Hren:
Filter orders that:
 - Have our Master/Author in their Master/Author/Blacklist blacklist.
 - Have their Master/Author in our Master/Author/Blacklist blacklist.

PR #940, opened 2018-06-04
refactor: makefile design

3Hren:
Now binaries produces by make contain the proper OS and ARCH suffixes even if they are compiled using cross-compilation feature.

Hence, there is no need for `build/*_linux` and `build/*_win32` targets. Now cross-platform binaries can be build using something like this:

`GOOS=windows GOARCH=386 make build/node`

You can read more about supported platforms and architectures: https://github.com/golang/go/wiki/MinimumRequirements.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,49 +1,65 @@
 #!/usr/bin/env make
-VER = v0.4.0
+
+# Version of the entire package. Do not forget to update this when it's time
+# to bump the version.
+VERSION = v0.4.0
+
+# Build tag. Useful to distinguish between same-version builds, but from
+# different commits.
 BUILD = $(shell git rev-parse --short HEAD)
-FULL_VER = $(VER)-$(BUILD)
 
-GOCMD=./cmd
-ifeq ($(GO), )
-    GO=go
+# Full version includes both semantic version and git ref.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,49 +1,65 @@
...
-LSGPU=${TARGETDIR}/lsgpu_$(OS_ARCH)
-PANDORA=${TARGETDIR}/pandora_$(OS_ARCH)
-OPTIMUS=${TARGETDIR}/sonmoptimus_$(OS_ARCH)
+WORKER     := ${TARGETDIR}/sonmworker_$(OS_ARCH)
+NODE       := ${TARGETDIR}/sonmnode_$(OS_ARCH)
+CLI        := ${TARGETDIR}/sonmcli_$(OS_ARCH)
+AUTOCLI    := ${TARGETDIR}/autocli_$(OS_ARCH)
+DWH        := ${TARGETDIR}/sonmdwh_$(OS_ARCH)
+RENDEZVOUS := ${TARGETDIR}/sonmrendezvous_$(OS_ARCH)
+RELAY      := ${TARGETDIR}/sonmrelay_$(OS_ARCH)
```
antmat: Reordered, but not sorted

3Hren: By significanceю

PR #946, opened 2018-06-05
refactor: use await with confirmations instead of old func

sokel:
Remove old await func for consistency.
New function works with confirmationa and returns receipt instead of one undefined transaction log.

PS: uniform any parts with awaiting for consistency


sokel:
changed

PR #956, opened 2018-06-06
feat: strip user path from binaries

3Hren:


antmat:
Oh my, LGTM! You're a hero!

3Hren:
No, I'm not!

I've tested on 3 platforms (linux, osx, windows) and it worked perfectly. But for unknown reasons seems like travis is agains us.

PR #961, opened 2018-06-07
fix: profile stats & updates

oopcode:
This PR fixes a bug that happens in the following situation:

* An order is created for an unknown profile (i.e., AuthorID/MasterID has no certificates), so ActiveAsks/ActiveBids is not incremented for that profile;
* A profile is created for that order's AuthorID/MasterID;
* Order's status changes to INACTIVE, we decrement the already known AuthorID/MasterID profile's ActiveAsks/ActiveBids and get a negative value.

The fix goes as follows: when creating a new profile, we count **active** asks and bids and store the new profile with relevant data.


insonmnia/dwh/server.go
```diff
@@ -391,9 +391,8 @@ func (m *DWH) watchMarketEvents() error {
 			}
 			m.processBlockBoundary(event)
 			dispatcher.Add(event)
-			if eventsCount < m.cfg.NumWorkers {
-				eventsCount++
-			} else {
+			eventsCount++
```
oopcode: With this change exactly `m.cfg.NumWorkers` events can be processed in parallel (not `m.cfg.NumWorkers + 1`, as in the old version).

insonmnia/dwh/server.go
```diff
@@ -801,8 +800,10 @@ func (m *DWH) onOrderPlaced(eventTS uint64, orderID *big.Int) error {
 		certificates, _ := json.Marshal([]*pb.Certificate{})
 		profile = &pb.Profile{UserID: order.AuthorID, Certificates: string(certificates)}
 	} else {
-		if err := m.updateProfileStats(conn, order.OrderType, userID, 1); err != nil {
-			return errors.Wrap(err, "failed to updateProfileStats")
+		if order.OrderStatus == pb.OrderStatus_ORDER_ACTIVE {
```
oopcode: Do not update stats for orders that come inactive but with a `DealID` set.

insonmnia/dwh/server.go
```diff
@@ -896,8 +898,10 @@ func (m *DWH) onOrderUpdated(orderID *big.Int) error {
 		}
 	}
 
-	if err := m.updateProfileStats(conn, marketOrder.OrderType, userID, -1); err != nil {
-		return errors.Wrapf(err, "failed to updateProfileStats (AuthorID: `%s`)", marketOrder.AuthorID.Unwrap().String())
+	if dwhOrder.GetOrder().OrderStatus == pb.OrderStatus_ORDER_ACTIVE {
```
oopcode: Only update stats for closed orders that we previously known to DWH as active.

insonmnia/dwh/sql.go
```diff
@@ -1609,27 +1612,38 @@ func (m *sqlStorage) newNetflagsWhere(builder squirrel.SelectBuilder, operator p
 
 func (m *sqlStorage) runQuery(conn queryConn, columns string, withCount bool, query string, args ...interface{}) (*sql.Rows, uint64, error) {
 	dataQuery := strings.Replace(query, "*", columns, 1)
+	var count uint64
```
oopcode: Querying for counts must be done before querying for data (because count rows should be closed before getting data rows).

PR #964, opened 2018-06-08
feat: reworked benchmarks in market

quasisamurai:
Now we can place Order with any count of benchmarks (but still not more than we support).
 In Deal now always stored the correct num of bencmarks (num that sonm supported when deal was opened)

### DIFF
--- a/blockchain/source/contracts/Market.sol
+++ b/blockchain/source/contracts/Market.sol
@@ -177,9 +177,9 @@ contract Market is Ownable {
         uint64[] _benchmarks
     ) public returns (uint){
 
-        require(_benchmarks.length == benchmarksQuantity);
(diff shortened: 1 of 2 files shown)


blockchain/source/contracts/Market.sol
```diff
@@ -713,6 +721,16 @@ contract Market is Ownable {
         }
     }
 
+    function ResizeBenchmarks(uint64[] _benchmarks) internal view returns (uint64[]) {
+        uint64[] memory benchmarks = new uint64[](benchmarksQuantity);
+        for(uint i = 0; i < benchmarksQuantity; i++){
```
antmat: just `i < _benchmarks.length` with no if ateward?

PR #966, opened 2018-06-08
fix: unwrap error

3Hren:


antmat:
in general it can not be nil, but ok

PR #971, opened 2018-06-09
feat: store netflags in dynamic array on market

quasisamurai:
- Now netflags stores in dynamic array.
- We can update num of netflags without both breaking API and redeploy market.
- CreateOrder accepts any count of netflags (less or eq we support).
- New event NumNetflagsUpdated.
-  New param in market constructor.

### DIFF
--- a/blockchain/source/contracts/Market.sol
+++ b/blockchain/source/contracts/Market.sol
@@ -77,7 +77,7 @@ contract Market is Ownable {
         address counterparty;
         uint duration;
         uint256 price;
-        bool[3] netflags;
+        bool[] netflags;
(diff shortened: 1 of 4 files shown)


blockchain/source/contracts/Market.sol
```diff
@@ -132,9 +133,12 @@ contract Market is Ownable {
 
     uint requestsAmount = 0;
 
-    // current length of benchmarks array (12)
+    // current length of benchmarks array (13)
```
antmat: I don't remember if it was changed

quasisamurai: lol, fixed

PR #972, opened 2018-06-09
fix: Optimus should now properly cut minimum RAM

3Hren:
This PR fixes a bug where a minimum number of some criteria (for example)RAM was too low than the minimum one by one. This is caused by loosing precision while converting to float64 and vice-versa - fixed by ceiling the result.

Also found and fixed a bug which caused the Optimus to overcut celling plans, because of invalid math - I used current benchmarks for calculating relational criteria, while lower bounds are calculated against total devices. Now we use initial benchmarks for this case.

### DIFF
--- a/optimus/devices_test.go
+++ b/optimus/devices_test.go
@@ -110,7 +110,7 @@ func TestConsumeCPU(t *testing.T) {
 	controller := gomock.NewController(t)
 	defer controller.Finish()
 
-	manager, err := newDeviceManager(devices, newMappingMock(controller))
+	manager, err := newDeviceManager(devices, devices, newMappingMock(controller))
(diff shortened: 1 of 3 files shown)


optimus/devices_test.go
```diff
@@ -467,3 +467,39 @@ func TestConsumeNetwork(t *testing.T) {
 	assert.Equal(t, uint64(95e6), manager.freeBenchmarks[5])
 	assert.Equal(t, uint64(10e6), manager.freeBenchmarks[6])
 }
+
+func TestConsumeRAMMin(t *testing.T) {
```
antmat: Ramin, brother!

PR #977, opened 2018-06-09
feat: Optimus now can filter by counterparty

3Hren:


optimus/optimus.go
```diff
@@ -229,6 +250,12 @@ func (m *workerControl) Execute(ctx context.Context) {
 			continue
 		}
 
+		// Ignore filled with counterparty orders that are not created for us.
```
oopcode: I'm not sure if this is relevant, but why do you filter orders client-side and do not use DWH filters? It looks like at least some of those filters are provide by DWH API.

3Hren: You mean program API? Please show how to do it, I will do.

PR #985, opened 2018-06-14
fix: move gasLimit into chainopts

sshaman1101:
This PR adds several files after the #976:
- `gasLimit` param added to chainOpts.
- Replaced constant usage with chainOpts param.
- Constants with defualt gas limit and gatekeeper addresses renamed using Sidechain\Masterchain.


sokel:
HUEVI PR

PR #986, opened 2018-06-14
feat: deposit-withdraw for node and cli

sshaman1101:
Chained with #985.

This PR adds able to transfer tokens between Masterchain and SONM sidechain.


insonmnia/node/tokens.go
```diff
@@ -38,6 +38,22 @@ func (t *tokenAPI) Balance(ctx context.Context, _ *sonm.Empty) (*sonm.BalanceRep
 	}, nil
 }
 
+func (t *tokenAPI) Deposit(ctx context.Context, amount *sonm.BigInt) (*sonm.Empty, error) {
+	if err := t.remotes.eth.MasterchainGate().PayIn(ctx, t.remotes.key, amount.Unwrap()); err != nil {
```
sokel: add token.approve() method here

insonmnia/node/tokens.go
```diff
@@ -38,6 +38,22 @@ func (t *tokenAPI) Balance(ctx context.Context, _ *sonm.Empty) (*sonm.BalanceRep
...
+		return nil, err
+	}
+
+	return &sonm.Empty{}, nil
+}
+
+func (t *tokenAPI) Withdraw(ctx context.Context, amount *sonm.BigInt) (*sonm.Empty, error) {
+	if err := t.remotes.eth.SidechainGate().PayIn(ctx, t.remotes.key, amount.Unwrap()); err != nil {
```
sokel: this too

insonmnia/node/tokens.go
sshaman1101: What value for `.approve()` is applicable here?

insonmnia/node/tokens.go
sshaman1101: ... and here?

blockchain/api.go
```diff
@@ -753,9 +753,29 @@ func (api *StandardTokenApi) AllowanceOf(ctx context.Context, from, to common.Ad
 	return api.tokenContract.Allowance(getCallOptions(ctx), from, to)
 }
 
-func (api *StandardTokenApi) Approve(ctx context.Context, key *ecdsa.PrivateKey, to common.Address, amount *big.Int) (*types.Transaction, error) {
+func (api *StandardTokenApi) Approve(ctx context.Context, key *ecdsa.PrivateKey, to common.Address, amount *big.Int) error {
 	opts := api.opts.getTxOpts(ctx, key, api.opts.gasLimit)
-	return api.tokenContract.Approve(opts, to, amount)
+	// reset allowance
+	tx, err := api.tokenContract.Approve(opts, to, big.NewInt(0))
```
sokel: nope =(

insonmnia/node/tokens.go
```diff
@@ -38,6 +41,57 @@ func (t *tokenAPI) Balance(ctx context.Context, _ *sonm.Empty) (*sonm.BalanceRep
...
+	currentAllowance, err := t.remotes.eth.LiveToken().AllowanceOf(ctx, me, blockchain.GatekeeperMasterchainAddr())
+	if err != nil {
+		return nil, errors.WithMessage(err, "cannot get current allowance")
+	}
+
+	if currentAllowance.Cmp(amount.Unwrap()) == -1 {
+		ctxlog.G(t.remotes.ctx).Debug("not enough allowance for deposit",
+			zap.String("amount", amount.Unwrap().String()),
+			zap.String("current", currentAllowance.String()))
+
```
sokel: reset allowance here.

blockchain/api.go
sshaman1101: Why? Setting allowance without resetting it to zero value doesn't work now, why not to perform this in `Approve()` method?

sokel: this not obvious behavior for method definition.
if you want, you may returning error, that allowance couldn't changed by this way.

sshaman1101: This makes API less obvious to use, and you ALWAYS need to reset allowance, so you will be forced to write a bit more **similar** code before **any** call of the `.Approve()` method. 

I'd prefer to describe this behavior in the method's docstrings and add extra handling for a case when you need to set allowance to zero.

sshaman1101:
rebased to the master and use Inc\Dec approval function from the #1010.
@sokel @3Hren @antmat @oopcode PTAL

insonmnia/node/tokens.go
```diff
@@ -38,6 +39,30 @@ func (t *tokenAPI) Balance(ctx context.Context, _ *sonm.Empty) (*sonm.BalanceRep
 	}, nil
 }
 
+func (t *tokenAPI) Deposit(ctx context.Context, amount *sonm.BigInt) (*sonm.Empty, error) {
+	if err := t.remotes.eth.MasterchainToken().IncreaseApproval(ctx, t.remotes.key, blockchain.GatekeeperMasterchainAddr(), amount.Unwrap()); err != nil {
```
antmat: either you have a bug here or docs of the increase approval are misleading. they say "approve should be called when current allowance = 0." (if it's true the API is still shit)

sshaman1101: The docs mean that an `.Approve()` method itself must be called with the current allowance set to zero, increase\decrease works well with any current approval value. 
I totally agree that method's docs describe the behavior not clearly, so I can fix it with an upcoming PR.

insonmnia/node/tokens.go
```diff
@@ -38,6 +39,30 @@ func (t *tokenAPI) Balance(ctx context.Context, _ *sonm.Empty) (*sonm.BalanceRep
...
+		return nil, errors.WithMessage(err, "cannot change allowance")
```
antmat: AFAIR errors.WithMessage also have quite strange formatting in attributes.

sshaman1101: No, it [doesn't](https://github.com/pkg/errors/blob/master/errors.go#L228), just a colon-separated string.

sshaman1101:
> looks like you missing payout calls on each side

No, `Payout` can be called only by owner (which is Gatekeeper in that case). And it performs payout calls when the PayIn called on any chain.

PR #987, opened 2018-06-14
Revert "fix: move gasLimit into chainopts"

sokel:
Reverts sonm-io/core#985


sokel:
We do not have a specified gas limit for chain options.
Gas limit must be specified for each method.

antmat:
It doesn't make any difference right now, so reverting is useless. Besides you are reverting #983 for some reasons

PR #989, opened 2018-06-14
feat: get profile level 

quasisamurai:

PR #991, opened 2018-06-14
fix: incorrect billing

quasisamurai:
This pr fixes a big bug while market refund least funds, that eq to blocked.balance before Bill. It also refunded to master, not to consumer 
chained:  #989


blockchain/source/contracts/Market.sol
```diff
@@ -415,7 +415,7 @@ contract Market is Ownable {
             } else {
                 emit Billed(dealID, deal.blockedBalance);
```
sokel: this is leverage event

PR #992, opened 2018-06-14
fix: proper cli login behaviour

sshaman1101:
This PR fixes two bugs:
- Now we do not mark the key as default if there is only one key is present into keystore.
- Now `sonmcli login` asking for pass-phrase, then trying to decrypt key and mark it as default only if decryption was successful.

### DIFF
--- a/cmd/cli/commands/login.go
+++ b/cmd/cli/commands/login.go
@@ -42,15 +42,15 @@ var loginCmd = &cobra.Command{
 				os.Exit(1)
 			}
 
-			// try to decrypt default key with given pass
+			// try to decrypt default key with given pass phrase
 			if _, err := ks.GetKeyWithPass(addr, pass); err != nil {
 				showError(cmd, "Cannot decrypt default key with given pass", err)
 				os.Exit(1)
 			}
 
-			// mark key as default if we can decrypt it with given
+			// mark key as default if we can decrypt it with given pass phrase
 			if err := ks.SetDefault(addr); err != nil {
-				cmd.Printf("Given address is not present into keystore.\r\nAvailable addresses:\r\n")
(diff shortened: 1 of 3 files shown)


cmd/cli/commands/login.go
```diff
@@ -56,6 +48,15 @@ var loginCmd = &cobra.Command{
 				os.Exit(1)
 			}
 
+			// mark key as default if we can decrypt it with given
```
antmat: given what?)

cmd/cli/commands/login.go
```diff
@@ -56,6 +48,15 @@ var loginCmd = &cobra.Command{
...
+			if err := ks.SetDefault(addr); err != nil {
+				cmd.Printf("Given address is not present into keystore.\r\nAvailable addresses:\r\n")
```
antmat: "into" means direction, not position.
I know that it was there previously, but as far as you move this code - better to fix it.

cmd/cli/commands/login.go
```diff
@@ -35,27 +35,28 @@ var loginCmd = &cobra.Command{
...
-			// try to decrypt default key with given pass
+			// try to decrypt default key with given pass phrase
 			if _, err := ks.GetKeyWithPass(addr, pass); err != nil {
 				showError(cmd, "Cannot decrypt default key with given pass", err)
 				os.Exit(1)
 			}
 
+			// mark key as default if we can decrypt it with given pass phrase
+			if err := ks.SetDefault(addr); err != nil {
+				cmd.Printf("Given address is not present in keystore.\r\nAvailable addresses:\r\n")
```
antmat: "in the keystore", but nevermind

PR #994, opened 2018-06-15
chore: unify method names (DWH)

oopcode:


oopcode:
it's based on squirrel's official examples, actually

PR #998, opened 2018-06-15
aux: redeploy contracts

sokel:
Redeploy `Market` and `ProfileRegistry` contracts


cmd/cli/commands/deals.go
```diff
@@ -11,14 +11,14 @@ import (
...
-	addToBlacklist    bool
+	blacklistTypeStr  string
 	crNewDurationFlag string
 	crNewPriceFlag    string
 )
 
 func init() {
 	dealListCmd.PersistentFlags().Uint64Var(&dealsSearchCount, "limit", 10, "Deals count to show")
-	dealCloseCmd.PersistentFlags().BoolVar(&addToBlacklist, "blacklist", false, "Add counterparty to blacklist")
+	dealCloseCmd.PersistentFlags().StringVar(&blacklistTypeStr, "blacklist", "none", "Whom to add to blacklist (worker, master or neither)")
```
sshaman1101: Properly describe valid flag values here: `neither` should be replaced with `none` as they parsed below.

cmd/cli/commands/deals.go
```diff
@@ -143,8 +143,8 @@ var dealOpenCmd = &cobra.Command{
 }
 
 var dealQuickBuyCmd = &cobra.Command{
-	Use:    "quick-buy <ask_id>",
-	Short:  "Copy given ASK order with BID type and open a deal with this orders",
+	Use:    "quick-buy <ask_id> [duration]",
```
sshaman1101: I'd prefer to implement this as the cobra flag with Duration type: less parsing, less checking for an args count, less description into a command (will be moved to the flag's help message).

proto/node.proto
```diff
@@ -64,12 +64,17 @@ service DealManagement {
     rpc CancelChangeRequest(BigInt) returns (Empty) {}
     // QuickBuy places BID order with the same parameters as given ASK order have,
     // then opens deal with this two orders.
-    rpc QuickBuy(BigInt) returns (Deal) {}
+    rpc QuickBuy(QuickBuyRequest) returns (Deal) {}
+}
+
+message QuickBuyRequest {
+    BigInt askId = 1;
```
sshaman1101: askId -> askID

blockchain/source/migrations/9_set_allowance_to_market.js
```diff
@@ -0,0 +1,14 @@
+var SNMD = artifacts.require('./SNMD.sol');
+var Market = artifacts.require('./Market.sol');
+
+module.exports = function (deployer, network) {
+    if (network === 'private') {
+        SNMD.at('0x26524b1234e361eb4e3ddf7600d41271620fcb0a')
+            .then(function (token) {
+                token.AddMarket(Market.address, { gasPrice: 0 });
```
sshaman1101: Looks weird, a file is called `set_allowance_to_market` but this call doesn't look like a setting of allowance.

blockchain/source/migrations/5_deploy_profileRegistry.js
```diff
@@ -1,7 +1,11 @@
 var ProfileRegistry = artifacts.require('./ProfileRegistry.sol');
 
 module.exports = function (deployer, network) {
-    if (network !== 'rinkeby') { // no reason to deploy to rinkeby
+    if (network === 'private') { // no reason to deploy to rinkeby
```
sshaman1101: This comment is misleading

cmd/cli/commands/deals.go
antmat: Thanks to our clear api we need to distinguish between 0 duration and it's absence. It would be quite dirty using the flag. Besides I find `sonmcli quickbuy 42 1h` more friendly than `sonmcli quickbuy --duration=1h 42`

blockchain/source/migrations/7_deploy_market.js
```diff
@@ -8,10 +8,11 @@ module.exports = function (deployer, network) {
...
         deployer.deploy(Market,
             '0x5db024c6469634f4b307135cb76e8e6806f007b3', // token address
-            Blacklist.address, // Blacklist address
-            OracleUSD.address, // Oracle address
-            '0xcc1cb65bdea124520dbdcc9e82b21b4352e45ad9', // ProfileRegistry address
+            '0x9ad1e969ec5842ee5d67414536813e224ceb56b1', // Blacklist address
+            '0x1f995e52dcbec7c0d00d45b8b1bf43b29dd12b5b', // Oracle address
+            '0xacfe1a688649fe9798b44a76b906fdca6e584a8d', // ProfileRegistry address
             12, // benchmarks quantity
+            3,
```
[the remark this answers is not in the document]
antmat: looks like it's netflag quantity, but comment would be helpfull here

cmd/cli/commands/deals.go
sshaman1101: But if the user agrees with proposed ASK duration it not necessary to explicitly set deal duration, we can handle it on the Node, like the change request API do.

PR #1008, opened 2018-06-17
refactor: contracts

sokel:
beauty some parts


blockchain/source/contracts/Market.sol
```diff
@@ -107,11 +107,10 @@ contract Market is Ownable {
     event NumBenchmarksUpdated(uint indexed newNum);
     event NumNetflagsUpdated(uint indexed newNum);
 
-    // due postgres/sqlite couldn't work w uint64
-    uint constant  maxBenchmarkValue  = 2 ** 63;
-
     // VARS
 
+    uint constant UINT64_MAX = 2 ** 63;
```
3Hren: It's not uint64 max.

PR #1010, opened 2018-06-18
feat: increase/decrease approval

sokel:
Add increase/decrease token functions.

PS: 1 line affected by `fmt`

### DIFF
--- a/blockchain/api.go
+++ b/blockchain/api.go
@@ -91,6 +91,14 @@ type TokenAPI interface {
 	AllowanceOf(ctx context.Context, from, to common.Address) (*big.Int, error)
 	// TotalSupply - all amount of emitted token
 	TotalSupply(ctx context.Context) (*big.Int, error)
+	// IncreaseApproval increase the amount of tokens that an owner allowed to a spender
+	// approve should be called when current allowance = 0. To increment
+	// allowed value is better to use this function to avoid 2 calls (and wait until the first transaction is mined)
+	IncreaseApproval(ctx context.Context, key *ecdsa.PrivateKey, spender common.Address, value *big.Int) error
+	// DecreaseApproval - decrease the amount of tokens that an owner allowed to a spender
+	// approve should be called when current allowance = 0. To decrement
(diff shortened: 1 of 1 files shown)


blockchain/api.go
```diff
@@ -755,6 +763,52 @@ func NewStandardToken(address common.Address, opts *chainOpts) (TokenAPI, error)
...
 }
 
+func (api *StandardTokenApi) IncreaseApproval(ctx context.Context, key *ecdsa.PrivateKey, spender common.Address, value *big.Int) error {
+	opts := api.opts.getTxOpts(ctx, key, api.opts.gasLimit)
+	tx, err := api.tokenContract.IncreaseApproval(opts, spender, value)
+	if err != nil {
+		return err
+	}
+
+	client, err := api.opts.getClient()
```
sshaman1101: You don't need to init client and check for error here, just use `api.client`.

blockchain/api.go
```diff
@@ -91,6 +91,14 @@ type TokenAPI interface {
 	AllowanceOf(ctx context.Context, from, to common.Address) (*big.Int, error)
 	// TotalSupply - all amount of emitted token
 	TotalSupply(ctx context.Context) (*big.Int, error)
+	// IncreaseApproval increase the amount of tokens that an owner allowed to a spender
+	// approve should be called when current allowance = 0. To increment
```
sshaman1101: What happens if I'll call this func with non-zero current approval?

blockchain/api.go
```diff
@@ -91,6 +91,14 @@ type TokenAPI interface {
...
+	// allowed value is better to use this function to avoid 2 calls (and wait until the first transaction is mined)
+	IncreaseApproval(ctx context.Context, key *ecdsa.PrivateKey, spender common.Address, value *big.Int) error
+	// DecreaseApproval - decrease the amount of tokens that an owner allowed to a spender
+	// approve should be called when current allowance = 0. To decrement
```
sshaman1101: Copy-paste?

blockchain/api.go
sokel: it should increased allowance to spender by sended value

PR #1013, opened 2018-06-18
feat: authentic masterchain token

quasisamurai:
Included:

- flattened contract code

- added ABI 

- added go_api


antmat:
Is it for abigen?

PR #1041, opened 2018-06-20
feat: testnet faucet contract

sokel:
This contract implement methods from SNMTToken, previously used to test at rinkeby.

Also, drop SNMTToken.


sokel:
thx, added

PR #1046, opened 2018-06-20
fix: relay check matching node

oopcode:
It might be the case that cluster reconfigures between Discover and Handshake; this PR checks that we only work with appropriate peers. It also adds dial retries for clients that experienced Discover/Handshake inconsistency.


oopcode:
ok ok

oopcode:
hmMMMmMmMmmmmmm.......

oopcode:
oh please review me
I’m so alone
in this repo
it hurts...

insonmnia/npp/relay/server.go
```diff
@@ -474,6 +475,20 @@ func (m *server) processHandshake(ctx context.Context, conn net.Conn, handshake
...
+	if !ok {
+		return errInvalidHandshake(errors.New("failed to get node by address"))
+	}
+
+	// Peer might have got a no longer valid node address while discovery.
+	if targetAddr != m.cfg.Addr.String() {
+		return errWrongNode()
+	}
+
+	if err := handshake.Validate(); err != nil {
```
3Hren: This is done few lines above.

insonmnia/npp/relay/client.go
```diff
@@ -188,8 +197,9 @@ func (m *client) Close() error {
 // dynamic DNS addition/removal. Thus, a single Relay endpoint as a hostname
 // should fit the best.
 type Dialer struct {
-	Addrs []string
-	Log   *zap.Logger
+	Addrs      []string
+	Log        *zap.Logger
+	NumRetries uint64
```
3Hren: Dead.

PR #1052, opened 2018-06-20
refactor: bind ask plan storage key to market address

antmat:
this PR adds market address to ask plan storage key to prevent situations when we migrate to new SC and have "hanging" ask plans in worker which we can not remove.


insonmnia/worker/salesman/salesman.go
```diff
@@ -69,10 +69,12 @@ func NewSalesman(opts ...Option) (*Salesman, error) {
 	if err := o.Validate(); err != nil {
 		return nil, err
 	}
+	marketAddr := blockchain.MarketAddr().Hex()
```
3Hren: Why to introduce this intermediate variable?

PR #1054, opened 2018-06-20
feat: specify worker settings in Optimus config

3Hren:
Now it is possible to define some worker specific settings in the autosell bot config.
For example now we can specify epoch interval, where the bot scans worker's resources and tries to sell them.

This is a BREAKING change.

### DIFF
--- a/etc/optimus.yaml
+++ b/etc/optimus.yaml
@@ -17,7 +17,7 @@ workers:
   0xccBe701e568577983E90968161ABB391759D589e@[::1]:15010:
     # Epoch is an interval of scanning worker's free resources and selling them
     # by cutting ask plans.
-    epoch: 5s
+    epoch: 60s
 
 benchmarks:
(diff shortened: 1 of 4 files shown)


etc/optimus.yaml
```diff
@@ -10,10 +10,14 @@ ethereum: &ethereum
...
   pass_phrase: "any"
 
-# List of workers trusted endpoints this bot manages.
+
+# Map of workers this bot manages defined by its trusted endpoints with their
+# settings.
 workers:
- - 0x8125721C2413d99a33E351e1F6Bb4e56b6b633FD@[::1]:15010
- # TODO: Support multiple timeouts per each worker. How?
+ 0xccBe701e568577983E90968161ABB391759D589e@[::1]:15010:
```
sshaman1101: looks like there should be **two** spaces

3Hren: Oh, indeed.

PR #1063, opened 2018-06-21
feat: dwh extended validators

oopcode:


insonmnia/dwh/sql.go
```diff
@@ -856,21 +856,48 @@ func (m *sqlStorage) InsertOrUpdateValidator(conn queryConn, validator *pb.Valid
 	}
 	defer rows.Close()
 
-	// Update if exists.
 	if rows.Next() {
-		// rows.Close is idempotent.
 		rows.Close()
```
sshaman1101: You're deferred row closing three lines above, isn't it?

oopcode: Yes, but:
1. It's idempotent,
2. In a transaction you **must** close your `Query()` rows before doing `Exec(); if you don't, postgres might fail to parse its own shit inside the connection.

PR #1066, opened 2018-06-22
refactor: flexible benchmarks

antmat:
This PR allows worker to perform and cache benchmarks separately for each device
and  enables parallelized benchmarks for GPU

### DIFF
--- a/insonmnia/worker/server.go
+++ b/insonmnia/worker/server.go
@@ -6,18 +6,20 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"reflect"
 	"strconv"
 	"sync"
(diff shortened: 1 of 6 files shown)


insonmnia/worker/server.go
```diff
@@ -1124,8 +1165,7 @@ func getDescriptionForBenchmark(b *pb.Benchmark) (Description, error) {
 		return Description{}, err
 	}
 	return Description{
-		autoremove: true,
```
sshaman1101: why drop autoremove?

antmat: because it's racy in terms of logs fetching

PR #1073, opened 2018-06-25
chore: refactor multisig tests

f4hrenh9it:


quasisamurai:
plz next time don't press update button

PR #1080, opened 2018-06-26
fix: deploy list contract

sokel:
fix bug with array length


sshaman1101:
Just curious: where is an abigen and address of the deployed contract?

sokel:
its not deployed yet.
Out of scope if this PR;

PR #1090, opened 2018-06-27
feat: non-negative LLS solver for Optimus

3Hren:
This adds a new more-physically-correct optimization method - the non-negative multidimensional linear least squares (NNLS).

It constraints the result vector values to be a non-negative, which in its turn means that each of them will show to the real benchmark price per basis.

Also this shows an extremely performance boost - million of orders are optimized in less than for a second.

However, currently this is an experimental proposal.

### DIFF
--- a/optimus/model_nnls.go
+++ b/optimus/model_nnls.go
@@ -1,0 +1,209 @@
+// This package implements of NNLS model, which solves the non-negative least
+// squares problem.
+//
+// See http://cmp.felk.cvut.cz/ftp/articles/franc/Franc-Hlavac-Navara-CAIP05.pdf
+// for more details.
+
+package optimus
+
+import (
+	"errors"
+	"fmt"
+	"math"
+	"strings"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+type SCAKKTModel struct {
+	MaxIterations int
+	Log           *zap.SugaredLogger
+}
+
+func newSCAKKTModel() modelFactory {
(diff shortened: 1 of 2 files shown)


3Hren:
image

optimus/model_nnls.go
```diff
@@ -0,0 +1,209 @@
...
+// Argument "A" represents input matrix. Each element of "A" must have the same
+// length. Vector "b" represents a measurement or output vector.
+// "Eps" is a tolerance for stopping iteration.
+//
+// The result tuple consists of three parameters. The first one returns
+// coefficients of the fitted linear function. It will have the same length
+// as elements of "A".
+// The second value is the number of iterations performed.
+// An error is returned if "A" and "b" are not the same length.
+func SCAKKT(A [][]float64, b []float64, eps float64, maxIterations int) ([]float64, int, error) {
```
montekki: BTW, what is considered a better practice in golang nowdays, using the multi-dim slices or using something like gonum for math-related computations?

3Hren: Currently all math-related packages in golang suck, unfortunately. Using `gonum` for matrix operations seems the best solution, however optimization engines we use require `[][]float64` instead of flattened matrix.
So the choice was dictated by the lowest level we use - [goml](https://github.com/cdipaolo/goml/blob/master/linear/linear.go#L133).

I personally dislike passing raw slices instead of using matrix abstraction, but while we using go, - we have almost no choice.

PR #1091, opened 2018-06-27
fix: Optimus will now set identity level

3Hren:


etc/optimus.yaml
```diff
@@ -49,6 +49,12 @@ workers:
     # Optional.
     # Default value is false.
     # dry_run: true
+    # Identity level specifies which kind of customers this bot allows to work
+    # with.
+    # Required.
+    # Possible values are: unknown, anonymous, registered, identified and
```
antmat: I don't think we should mention `unknown` in comments. Moreover - check for non-zero value is a good idea

antmat: Ah, it's already required, so should not be zero. Then remove `unknown` from comment please

etc/optimus.yaml
```diff
@@ -49,6 +49,12 @@ workers:
     # Optional.
     # Default value is false.
     # dry_run: true
+    # Identity level specifies which kind of customers this bot allows to work
+    # with.
+    # Required.
+    # Possible values are: anonymous, registered, identified and
+    # professional.
```
antmat: just curious - what's the purpose of newline here?

3Hren: 80 margin limit

PR #1093, opened 2018-06-27
refactor: better logs

antmat:
This PR adds logs for each resource pool operation, providing better understanding what's going on.


insonmnia/resource/pool.go
```diff
@@ -104,7 +104,14 @@ func (m *Scheduler) Consume(askPlan *sonm.AskPlan) error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	m.askPlanPools[askPlan.ID] = newPool(askPlan.Resources)
-	return m.pool.consume(askPlan.ID, askPlan.Resources)
+
+	if err := m.pool.consume(askPlan.ID, askPlan.Resources); err != nil {
+		err = fmt.Errorf("failed to consume resources for ask plan %s: %s", askPlan.ID, err)
```
3Hren: Here and below: I'm not the fan of this idea. Anyway the error should be logged at the upper level.

PR #1102, opened 2018-06-27
fix: proper check for worker availability for the deal

sshaman1101:
This commit adds a check that worker is available
when the user making deal with `quick-buy` handle.

Also, this commit remove a message that warns the user about the
worker inaccessibility when showing deal list.


cmd/cli/commands/printers.go
```diff
@@ -261,7 +261,7 @@ func printDealsList(cmd *cobra.Command, deals []*pb.Deal) {
 
 }
 
-func printDealInfo(cmd *cobra.Command, info *pb.DealInfoReply, changes *pb.DealChangeRequestsReply) {
+func printDealInfo(cmd *cobra.Command, info *pb.DealInfoReply, changes *pb.DealChangeRequestsReply, suppressWarning bool) {
```
3Hren: I'd rather replaced this with human-readable flags.
Something liks:

```go
type printFlags int
func (m printFlags) HasSuppressWarnings() bool { return int(m) & suppressWarningsFlag }
const suppressWarningsFlag = 0x01
func printDealInfo(cmd *cobra.Command, info *pb.DealInfoReply, changes *pb.DealChangeRequestsReply, flags printFlags) {
    if flags.HasSuppressWarnings() {...}
}
```

PR #1104, opened 2018-06-28
refactor: extend NPP identifier with protocol

3Hren:
This allows to specify some protocol you want to publish/resolve as the Rendezvous allows. It is required for further NAT traversal extensions.

What is changed: now relay protocol allows to specify "protocol" field in order to match between different protocols for the same peer ID. In both Relay and Rendezvous meeting groups are now associated with complex string representation a-la URI.

This is the BREAKING change. Both Rendezvous, Worker and Node must be updated. Relay, however, does not need to.

Output example:
```json
{
  "state": {
    "tcp://0xccBe701e568577983E90968161ABB391759D589e": {
      "servers": {
        "74b76170-2226-4f28-bc61-b9f5b13e44bc": {
          "publicAddr": {
            "protocol": "tcp",
            "addr": {
              "addr": "46.148.198.134",
              "port": 52791
            }
          },
          "privateAddrs": [
            {
              "protocol": "tcp",
              "addr": {
                "addr": "172.16.1.105",
                "port": 52791
              }
            }
          ]
        }
      }
    }
  }
}
```

Note that `tcp://` added.


insonmnia/npp/relay/metrics.go
```diff
@@ -46,7 +46,7 @@ func (m *metrics) Dump() *sonm.RelayMetrics {
 	netMetrics := map[string]*sonm.NetMetrics{}
 
 	for addr, metrics := range m.net {
-		netMetrics[addr.Hex()] = &sonm.NetMetrics{
+		netMetrics[addr.String()] = &sonm.NetMetrics{
```
antmat: Y?

PR #1110, opened 2018-06-28
feat: contract registry

antmat:
This PR replaces static smart contract addresses with contract registry smart contract, whichh is loaded on start and caches all addresses.

### DIFF
--- a/blockchain/api.go
+++ b/blockchain/api.go
@@ -171,7 +171,7 @@ type BasicAPI struct {
 	sidechainGate    SimpleGatekeeperAPI
 }
 
-func NewAPI(opts ...Option) (API, error) {
+func NewAPI(ctx context.Context, opts ...Option) (API, error) {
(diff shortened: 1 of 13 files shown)


blockchain/api.go
```diff
@@ -267,11 +351,129 @@ func (api *BasicAPI) SidechainGate() SimpleGatekeeperAPI {
...
+	oracleUsdAddress             common.Address
+	gatekeeperMasterchainAddress common.Address
+	gatekeeperSidechainAddress   common.Address
+	testnetFaucetAddress         common.Address
+
+	registryContract *marketAPI.AddressHashMap
+}
+
+func registryKey(key string) [32]byte {
+	if len(key) > 32 {
```
3Hren: Isn't it must be equal to 32?

antmat: no, it is at most 32 byte

PR #1112, opened 2018-06-28
fix: close events chan (blockchain api)

oopcode:


3Hren:
Why?

oopcode:
writers must close their channnels before exiting, this is an expected api behavior

PR #1116, opened 2018-06-28
fix: default live endpoints

antmat:
migrating to live


3Hren:
Can we still checkout to testnet using #1110?

antmat:
yep, via config

3Hren:
Ok, LGTM.

PR #1134, opened 2018-06-29
feat: add ask plan timestamps

antmat:
This PR adds creation timestamp orderPlaced timestamp and function to get time for which order has not been sold, which can be usefull for optimus


proto/timestamp.go
```diff
@@ -4,6 +4,31 @@ import "time"
 
 // Unix returns the local time.Time corresponding to the given Unix time
 // since January 1, 1970 UTC.
-func (m Timestamp) Unix() time.Time {
```
3Hren: Why it's required to change `this` from value to pointer?

proto/ask_plan.go
```diff
@@ -60,6 +61,16 @@ func (m *AskPlan) Validate() error {
 	return m.GetResources().GetGPU().Validate()
 }
 
+func (m *AskPlan) SaleTime() time.Duration {
```
3Hren: It's not **time** actually, but the duration.

antmat: yep, I'll rename, but this one is exactly you are looking for.

proto/timestamp.go
antmat: to be consistent and safely call Unix() on nil receiver

proto/ask_plan.go
3Hren: No, it's not, because if I create an ask plan and if it never be placed on the market, it returns 0, while must return `birthTime - now`.

antmat:
I think you should not react to temporary problems with blockchain, because situation when order can not be placed in blockchain is not normal and may require manual intervention.

antmat:
vice versa — returning `birthTime - now` when deal was closed several seconds ago may introduce races when it was actually in deal several seconds ago, but you will see huge unsold time

proto/timestamp.go
3Hren: Consistent with what? For example in `std` `func (t Time) Unix() int64`. Still did not get the idea.

> safely call Unix() on nil receiver

I still doubt that it's the best practice. For example from [this](https://github.com/golang/go/wiki/CodeReviewComments#receiver-type):

> If the receiver is a small array or struct that is naturally a value type (for instance, something like the time.Time type), with no mutable fields and no pointers, or is just a simple basic type such as int or string, a value receiver makes sense.
A value receiver can reduce the amount of garbage that can be generated; if a value is passed to a value method, an on-stack copy can be used instead of allocating on the heap. (The compiler tries to be smart about avoiding this allocation, but it can't always succeed.) Don't choose a value receiver type for this reason without profiling first.

PR #1135, opened 2018-06-29
fix: transaction like consumption in Optimus

3Hren:
When combining orders into a device manager, sometimes, when further benchmarks did not fit in the set, Optimus stole those resources, making all following attempts to fail with resource exhausted error.
This commit fixes that behaviour.

Also as a minor optimization fast GPU filtering by memory was introduced.

### DIFF
--- a/optimus/devices.go
+++ b/optimus/devices.go
@@ -184,6 +184,21 @@ func newDeviceManager(devices *sonm.DevicesReply, freeDevices *sonm.DevicesReply
 }
 
 func (m *DeviceManager) Consume(benchmarks sonm.Benchmarks) (*sonm.AskPlanResources, error) {
+	// Transaction-like resource restoring while consuming in case of errors.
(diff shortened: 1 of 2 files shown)


optimus/devices.go
```diff
@@ -337,11 +352,28 @@ func (m *DeviceManager) consumeGPU(minCount uint64, benchmarks []uint64) (*sonm.
...
 
+			// Fast filter by GPU memory benchmark.
+			// All GPUs in the subset must have at least(!) the required memory
+			// number.
+			for _, gpu := range subset {
+				for id := range currentBenchmarks {
+					if m.mapping.SplittingAlgorithm(id) == sonm.SplittingAlgorithm_MIN {
+						if benchmark, ok := gpu.Benchmarks[uint64(id)]; ok {
+							if currentBenchmarks[id] > benchmark.Result {
+								continue subsetLoop
```
antmat: consider using break and subfunctions instead of labels

PR #1152, opened 2018-06-30
proposal: better error handling for the cli

sshaman1101:
This commit changes CLI behavior in the following way:
1. Using RunE instead of Run for each cobra command.
2. Setting SilenceUsage = true, if not - cli will print help usage on every error.
3. Setting SilenceErrors = true, if not - cli will show their own error and then handle out custom code that controls the os.Exit status.
4. Introducing `newCmd` and `newCmdWithKeys` wrappers. The first one is used to wrapping cobra.Command with `SilenceUsage` and  `SilenceErrors` flags. The second one is used to wrapping command with the key loading code (if required and not handled externally).
5. Removed all of `os.Exit(1)` calls, excepts only one that handles single point of error handling now.

### DIFF
--- a/cmd/cli/commands/blacklist.go
+++ b/cmd/cli/commands/blacklist.go
@@ -1,7 +1,7 @@
 package commands
 
 import (
-	"os"
+	"fmt"
 
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/sonm-io/core/proto"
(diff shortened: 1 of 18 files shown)


3Hren:
Good. Do more.

cmd/cli/commands/blacklist.go
```diff
@@ -17,71 +17,70 @@ func init() {
 }
 
 var blacklistRootCmd = &cobra.Command{
-	Use:   "blacklist",
-	Short: "Manage blacklisted addresses",
+	Use:                "blacklist",
+	Short:              "Manage blacklisted addresses",
+	PersistentPostRunE: loadKeyStoreIfRequired,
```
antmat: img

cmd/cli/commands/master.go
```diff
@@ -37,113 +37,124 @@ var masterListCmd = &cobra.Command{
...
 
 	_, err = mm.WorkerRemove(ctx, &pb.WorkerRemoveRequest{
 		Master: pb.NewEthAddress(master),
 		Worker: pb.NewEthAddress(worker),
 	})
 	if err != nil {
-		showError(cmd, "Cannot drop master - worker relationship", err)
-		os.Exit(1)
+		return fmt.Errorf("cannot drop master-worker relationship: %v", err)
 	}
```
antmat: Y U NO SHOW OK??

PR #1158, opened 2018-07-03
fix: incremental logs (blockchain api)

oopcode:
A situation was possible where you call `EventsApi.GetEvents()` and run out of memory because all the logs are loaded at the same time. This PR fixes the above issue by collecting logs in batches (batch size is configurable).


blockchain/api.go
```diff
@@ -1364,24 +1369,43 @@ func (api *BasicEventsAPI) GetLastBlock(ctx context.Context) (uint64, error) {
 
 func (api *BasicEventsAPI) GetEvents(ctx context.Context, filter *EventFilter) (chan *Event, error) {
 	var out = make(chan *Event, 128)
+	// Number of blocks already present in blockchain that we will read in batches.
+	lastBlock, err := api.GetLastBlock(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get latest block number: %v", err)
+	}
+
+	if filter.FromBlock.Uint64() > lastBlock {
```
antmat: why not just wait?

blockchain/api.go
```diff
@@ -69,8 +69,8 @@ type ProfileRegistryAPI interface {
 type EventsAPI interface {
 	GetEvents(ctx context.Context, filter *EventFilter) (chan *Event, error)
 	GetLastBlock(ctx context.Context) (uint64, error)
-	GetMarketFilter(fromBlock *big.Int) *EventFilter
-	GetMultiSigFilter(addresses []common.Address, fromBlock *big.Int) *EventFilter
+	GetMarketFilter(fromBlock *big.Int, blocksBatchSize int64) *EventFilter
```
antmat: Looks like it's not the parametere of call, but configuration parameter of EventsAPI

blockchain/config.go
```diff
@@ -37,6 +38,7 @@ func (m *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		MasterchainEndpoint  string         `yaml:"endpoint"`
 		SidechainEndpoint    string         `yaml:"sidechain_endpoint"`
 		ContractRegistryAddr common.Address `yaml:"contract_registry"`
+		BlocksBatchSize      int64          `yaml:"blocks_batch_size" default:"200000"`
```
antmat: did you test that this value is not too high to be default?

blockchain/api.go
```diff
@@ -1364,24 +1369,43 @@ func (api *BasicEventsAPI) GetLastBlock(ctx context.Context) (uint64, error) {
...
+			toBlock            = big.NewInt(0)
+			tk                 = time.NewTicker(time.Millisecond * 100)
 		)
 		for {
 			select {
 			case <-ctx.Done():
 				close(out)
 				return
 			case <-tk.C:
+				if blocksLeft > 0 {
```
antmat: rather complex logic, why not just update toBlock constantly? does handle return error if toBlock is larger than known block?

blockchain/api.go
oopcode: because it's weird to wait for _something in the future_, not going to change

blockchain/api.go
oopcode: ok will do

PR #1162, opened 2018-07-04
feat: let cli read it's config from --config flag

montekki:
Might be useful for debugging when we have two or more nodes on one host, so we don't have to re-login all the time, can specify node addreses in config etc.

### DIFF
--- a/cmd/cli/commands/common.go
+++ b/cmd/cli/commands/common.go
@@ -44,6 +44,7 @@ var (
 	timeoutFlag     = 60 * time.Second
 	insecureFlag    bool
 	keystoreFlag    string
+	configFlag      string
 
 	// logging flag vars
 	logType       string
@@ -55,7 +56,7 @@ var (
 	prependStream bool
 
 	// session-related vars
-	cfg      *config.Config
+	cfg      = &config.Config{}
 	creds    credentials.TransportCredentials
 	keystore *accounts.MultiKeystore
 )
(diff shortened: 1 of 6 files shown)


cmd/cli/config/config.go
```diff
@@ -103,6 +103,15 @@ func getConfigPath(p ...string) (string, error) {
 		}
 	}
 
-	cfgPath = path.Join(cfgPath, configName)
+	info, err := os.Stat(cfgPath)
```
sshaman1101: we have the `util.DirectoryExists()` utility, so this checks can be replaced with one line

cmd/cli/commands/common.go
```diff
@@ -83,20 +84,30 @@ func getDefaultKey() (*ecdsa.PrivateKey, error) {
 }
 
 func init() {
+	cobra.OnInitialize(func() {
+		var err error
+		cfg, err = config.NewConfig(configFlag)
+		fmt.Println("Loading config ", configFlag)
```
sshaman1101: Looks like a forgotten debug print

cmd/cli/config/config.go
```diff
@@ -103,6 +103,11 @@ func getConfigPath(p ...string) (string, error) {
...
 
-	cfgPath = path.Join(cfgPath, configName)
-	return cfgPath, nil
+	switch {
+	case util.FileExists(cfgPath) == nil:
+		return cfgPath, nil
+	case util.DirectoryExists(cfgPath) == nil:
+		return path.Join(cfgPath, configName), nil
+	}
+	return "", fmt.Errorf("No such file or directory: %s", cfgPath)
```
sshaman1101: chained errors should be lowercased according to [Golang's wiki](https://github.com/golang/go/wiki/Errors)

PR #1165, opened 2018-07-04
refactor: Optimus decomposition

3Hren:
Two-way optimization algorithms introduced here. Its main aim is to
reduce false-positive ask-plan creation. One way to achieve this is to
do optimization for both virtual devices and for natural ones.

Preparing worker input such as fetching its devices and ask-plans are now
performed concurrently to save some time. Its duration is limited using
new option: "prelude_timeout".

Fetching orders from the DWH is now performed on-demand, however caching
still exists.

#### How to review this?
1. Type "I have no idea what's going on here".
2. Approve.


optimus/optimus.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
```
sshaman1101: si señor

PR #1167, opened 2018-07-04
fix: dwh proper exit

oopcode:
When DWH is stopped and restarted, it can produce errors due to attempts to re-process older events. This PR fixes this by doing the following:

1. DWH saves **full** information (block number, transaction index, index of log in receipt) about last processed event (not just block number).
2. DWH checks that currently processed event has not been processed before.
3. Last event info is saved to DB **only** after this batch has been fully processed.
4. `DWH.Stop()` (called after a `sigint`, for example) cannot interrupt processing current batch of events.

This guarantees that there's no inconsistency between what DWH *thinks* is the last processed event and what really has been processed.


insonmnia/dwh/sql.go
```diff
@@ -1116,6 +1085,51 @@ func (m *sqlStorage) CheckStaleID(conn queryConn, id *big.Int, entity string) (b
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to GetLastEvent: %v", err)
+	}
+	defer rows.Close()
+
+	if ok := rows.Next(); !ok {
+		return nil, errors.New("GetLastEvent: no entries")
+	}
+
+	var (
```
montekki: BTW why do you declare vars of the same type this way instead of var a, b, c uint64?

blockchain/types.go
```diff
@@ -9,9 +9,29 @@ import (
...
 	Data        interface{}
 	BlockNumber uint64
+	TxIndex     uint64
+	Index       uint64
 	TS          uint64
 }
 
+func (m *Event) PrecedesOrEquals(other *Event) bool {
+	var (
+		byBlockNumber = func(x, y *Event) bool { return x.BlockNumber < y.BlockNumber }
```
montekki: Why don't you create a slice of functions so you don't have to repeat their names in the range?

```go
	a := []func(int){func (a int) { fmt.Println(a + 1) }, func (a int) { fmt.Println(a + 2) }}
	for _, f := range a {
		f(1)
	}
```

oopcode: good point

PR #1169, opened 2018-07-05
chore: dwh startup refactoring

oopcode:
prettify locks for `serveGRPC()` and `serveHttp()`.

notice: locking these sections is required to properly handle interrupts while startup.

### DIFF
--- a/etc/dwh.yaml
+++ b/etc/dwh.yaml
@@ -18,11 +18,6 @@ blockchain:
   contract_registry: "0xaf1ffd7f652be7e9a0854a42b2d3046f853f80f1"
   blocks_batch_size: 500000
 
-#blockchain:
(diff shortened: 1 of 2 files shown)


etc/dwh.yaml
```diff
@@ -20,10 +20,14 @@ ethereum:
 blockchain:
   endpoint: "https://rinkeby.infura.io/00iTrs5PIy0uGODwcsrb"
   sidechain_endpoint: "http://localhost:8545"
-  #  sidechain_endpoint: "https://sidechain-dev.sonm.com"
   contract_registry: "0xaf1ffd7f652be7e9a0854a42b2d3046f853f80f1"
   blocks_batch_size: 500000
 
+#blockchain:
```
3Hren: Duplicated section.

PR #1191, opened 2018-07-12
feat: listener wrapper with context

3Hren:


3Hren:
Unfortunately seems it has no life. So closing for now.

PR #1195, opened 2018-07-12
feat: make masterchain gas price configurable

sokel:
This option now usefull for `deposit` and gatekeeper `withdraw`;
If option not defined gas price is determined by the latest blocks median gas price.

### DIFF
--- a/blockchain/config.go
+++ b/blockchain/config.go
@@ -41,7 +41,7 @@ func (m *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		SidechainEndpoint    string         `yaml:"sidechain_endpoint"`
 		ContractRegistryAddr common.Address `yaml:"contract_registry"`
 		BlocksBatchSize      uint64         `yaml:"blocks_batch_size" default:"500"`
-		MasterchainGasPrice  string         `yaml:"masterchain_gas_price"`
+		MasterchainGasPrice  GasPrice       `yaml:"masterchain_gas_price"`
 	}
 
 	if err := unmarshal(&cfg); err != nil {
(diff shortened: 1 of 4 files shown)


blockchain/config.go
```diff
@@ -13,6 +13,7 @@ type Config struct {
 	SidechainEndpoint    url.URL
 	ContractRegistryAddr common.Address
 	BlocksBatchSize      uint64
+	MasterchainGasPrice  int64
```
3Hren: What if it is negative?

etc/node.yaml
```diff
@@ -51,3 +51,9 @@ ethereum:
   pass_phrase: "any"
 
 metrics_listen_addr: "127.0.0.1:14003"
+
+
+blockchain:
+  # gas price for masterchain transactions, in wei
+  # default is 20 Gwei
+  masterchain_gas_price: 20000000000
```
3Hren: Wouldn't it better to write directly "20Gwei" instead, like we do in duration fields?

blockchain/options.go
```diff
@@ -63,14 +63,14 @@ type options struct {
...
-			gasPrice:           defaultMasterchainGasPrice,
+			gasPrice:           big.NewInt(0).SetUint64(defaultMasterchainGasPrice),
 			gasLimit:           defaultMasterchainGasLimit,
 			endpoint:           defaultMasterchainEndpoint,
 			logParsePeriod:     defaultLogParsePeriod,
 			blockConfirmations: defaultBlockConfirmations,
 		},
 		sidechain: &chainOpts{
-			gasPrice:           defaultSidechainGasPrice,
+			gasPrice:           big.NewInt(0).SetUint64(defaultSidechainGasPrice),
```
sshaman1101: big.NewInt(defaultSidechainGasPrice)

sokel: fixed

PR #1196, opened 2018-07-12
feat: ping worker before deal

sshaman1101:
Chained with #1193.

Now node trying to connect to the worker before opening deal
(both OpenDeal and QuickBuy). If a worker is not responding
to the status request - deal will not be opened and user
will be noticed about worker unavailability. User can force
deal opening with the `--force` flag passed to related CLI's sub-commands.


insonmnia/node/deals.go
```diff
@@ -97,17 +115,30 @@ func (d *dealsAPI) Open(ctx context.Context, req *pb.OpenDealRequest) (*pb.Deal,
...
-		if err != nil {
-			return nil, fmt.Errorf("failed to fetch ask order for duration lookup: %s", err)
-		}
 		duration = ask.Duration
 	} else {
 		duration = uint64(req.GetDuration().Unwrap().Seconds())
 	}
 
+	if !req.Force {
+		// check that worker is available
```
3Hren: Unnecessary comment, since log below says the same.

insonmnia/node/server.go
```diff
@@ -69,10 +69,28 @@ func (re *remoteOptions) getWorkerClientForDeal(ctx context.Context, id string)
...
+func (re *remoteOptions) isWorkerAvailable(ctx context.Context, addr common.Address) bool {
+	worker, closer, err := re.getWorkerClientByEthAddr(ctx, addr.Hex())
+	if err != nil {
+		return false
+	}
+
+	defer closer.Close()
+
+	_, err = worker.Status(ctx, &pb.Empty{})
+	if err != nil {
```
3Hren: return err == nil

3Hren:
Conflicts.

PR #1202, opened 2018-07-13
refactor: complete Node refactoring

3Hren:
The goal of this commit is refactoring Node source code to achieve
logical decomposition.

Fixed race conditions while failed to start some parts of the Node
servers. For example when REST server failed to start there was resource
leakage.

Main function now clearly describes that it waits either for
interruption or when the Node stopped itself, for example when it fails
to start properly.

Simplified cerficication creation by wrapping into simplified API that
accepts the context and returns transport credentials. The certificate
rotation will be stopped when the context is expired.

Node no longer keeps the Context variable as a struct field. Doing so is
wrong, because there is no "logical" context in such variable. It was
used only to extract logger instance, which is done now directly via
member field.
At the opposite, functions that do some context-specific things, like
serving listeners etc. now accept Context and are terminated when that
context is canceled.

Node server is now configured using options pattern to be able to
activate or deactivate components, like gRPC and REST servers, its
authentication, logging etc.


cmd/node/main.go
```diff
@@ -30,29 +28,26 @@ func run() error {
...
+	log := logging.BuildLogger(cfg.Log.LogLevel())
+	ctx := ctxlog.WithLogger(context.Background(), log)
 
-	key, err := cfg.Eth.LoadKey()
-	if err != nil {
-		return fmt.Errorf("failed to load Ethereum keys: %s", err)
-	}
-
-	n, err := node.New(ctx, cfg, key)
+	server, err := node.New(ctx, cfg, node.WithLog(log))
```
antmat: use wg ctx

util/defergroup/mod.go
```diff
@@ -0,0 +1,50 @@
...
+type DeferGroup struct {
+	fn       []func()
+	canceled bool
+}
+
+func (m *DeferGroup) Defer(fn func()) {
+	m.fn = append(m.fn, fn)
+}
+
+func (m *DeferGroup) Destroy() {
```
antmat: I think Exec() will be better

util/defergroup/mod.go
```diff
@@ -0,0 +1,50 @@
...
+	if m.canceled {
+		return
+	}
+
+	for _, fn := range m.reversed() {
+		fn()
+	}
+}
+
+func (m *DeferGroup) Ok() {
```
antmat: Cancel() because it cancels execution and sets cancelled variable to true

util/defergroup/mod.go
```diff
@@ -0,0 +1,50 @@
...
+	m.canceled = true
+}
+
+func (m *DeferGroup) reversed() []func() {
```
antmat: cmon, can't you just iterate from end by indexes without copy?

PR #1203, opened 2018-07-17
feat: show Ethereum balance on live network

sshaman1101:
This commit extends the Ethereum API Client with `GetEthereumBalanceAt` method
that allows to obtain Ethereum balance. Now `TokenAPI` returns an `Balance`
struct which represents balances in both Eth and SNM tokens.
Thus protobuf's API extended to be able to represent both values.
And finally, CLI able to print balance also in Ethereum (for Masterchain).


cmd/cli/commands/printers.go
```diff
@@ -402,9 +402,10 @@ func printTaskStart(cmd *cobra.Command, start *pb.StartTaskReply) {
 func printBalanceInfo(cmd *cobra.Command, reply *pb.BalanceReply) {
 	side := reply.GetSideBalance().ToPriceString()
 	live := reply.GetLiveBalance().ToPriceString()
+	eth := reply.GetLiveEthBalance().ToPriceString()
```
3Hren: I'd rather named it as `ethLive`. Just for clarity.

PR #1218, opened 2018-07-20
feat: update deals before removal (DWH)

oopcode:
A weird request came from the guys that develop our telegram notification bot: 

> Please update deals before deleting them.

Looks like they subscribe to postgres events and need deals final state to be written to DB.


3Hren:
Hmm, looks mad.
Is there someone who else can comment this? Otherwise this is the candidate for closing.

PR #1221, opened 2018-07-23
feat: restore benchmarks from map

sshaman1101:


3Hren:
I don't like such hardcode in common marketplace API, sorry. We already [have](https://github.com/sonm-io/benchmarks-list/blob/master/list.json) benchmark list, where we go when it's required to map benchmarks by name/id/device/algorithm etc.

PR #1255, opened 2018-08-02
feat: enable pprof in Optimus

3Hren:


3Hren:
This helped to find and optimize #1256.

PR #1266, opened 2018-08-06
feat: BTFS volumes

3Hren:


3Hren:
Native solution seems preferred, but currently there are problems with fuse binding. So this PR is replaced with #1274.

PR #1272, opened 2018-08-06
aux: enable pprof for Node

sshaman1101:


sshaman1101:
Oh, I'd completely missed that we already have a pretty helper for the pprof server. Fixed!

PR #1274, opened 2018-08-06
feat: BTFS volume plugin for Worker

3Hren:
This commit activates an ability to spawn SONM tasks with torrent
filesystem attached as a volume.

Internally this is achieved by using BTFS application that constructs a
torrent client and mounts itself on a filesystem via fuse. It's the
read-only solution designed to distribute large amount of data across
multiple task without single point of failure - all tasks help each
other to share the data, which reduces network pressure and improves
application scalability.
The contents of the files will be downloaded on-demand as they are read 
by applications. Tools like ls, cat and cp works as expected.

[asciicast](https://asciinema.org/a/5ZIETLiht4r5SHRxZ19EnQV8n)

### DIFF
--- a/etc/worker.yaml
+++ b/etc/worker.yaml
@@ -84,6 +84,7 @@ plugins:
     root: /var/lib/docker-volumes
     drivers:
 #      cifs: {}
+#      btfs: {}
 
   overlay:
     drivers:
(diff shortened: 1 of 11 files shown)


noxiouz:
Probably I read it wrong, but do we need some mechanism to failed mounts from previous launch if they still present? Or is it impossible to end up with orphaned mounts? Otherwise, LGTM

3Hren:
Yes, you are right! I also noticed that. However this is the common volume-specific problem - CIFS also suffers from dead-mounts. The solution requires integration with salesman and teaching volumes under what deal they are created, and with current API it's impossible to achieve this without refactoring. So I left it as is with keeping in mind that it should be fixed in the next PR under refactoring scope.

PR #1281, opened 2018-08-07
refactor: simplify task state storing and restoring

montekki:


insonmnia/worker/server.go
```diff
@@ -371,18 +372,38 @@ func (m *Worker) cancelDealTasks(deal *pb.Deal) error {
...
-	Spec        pb.TaskSpec   `json:"spec,omitempty"`
+	Description descriptionMarshallable   `json:"description,omitempty"`
+	Cinfo       containerInfoMarshallable `json:"cinfo,omitempty"`
+	Spec        pb.TaskSpec               `json:"spec,omitempty"`
+}
+
+type descriptionMarshallable struct {
+	Description
+	Reference reference.Field `json:"Reference"`
+	Networks  []*structs.NetworkSpec
```
antmat: why can't we just make networks in description public?

insonmnia/worker/server.go
```diff
@@ -371,18 +372,38 @@ func (m *Worker) cancelDealTasks(deal *pb.Deal) error {
...
-	Cinfo       ContainerInfo `json:"cinfo,omitempty"`
-	Spec        pb.TaskSpec   `json:"spec,omitempty"`
+	Description descriptionMarshallable   `json:"description,omitempty"`
+	Cinfo       containerInfoMarshallable `json:"cinfo,omitempty"`
+	Spec        pb.TaskSpec               `json:"spec,omitempty"`
+}
+
+type descriptionMarshallable struct {
+	Description
+	Reference reference.Field `json:"Reference"`
```
antmat: maybe just store reference.Field in description for simplicity?

insonmnia/worker/server.go
```diff
@@ -371,18 +372,38 @@ func (m *Worker) cancelDealTasks(deal *pb.Deal) error {
...
+	Networks  []*structs.NetworkSpec
+}
+
+type containerInfoMarshallable struct {
```
antmat: why can't we introduce wrapper for Public key instead of containerInfo?

insonmnia/worker/server.go
```diff
@@ -949,6 +970,11 @@ func (m *Worker) setupRunningContainers() error {
 				continue
 			}
 
+			info.Cinfo.ContainerInfo.PublicKey, _ = ssh.ParsePublicKey(info.Cinfo.PublicKey)
+
+			info.Description.Description.networks = info.Description.Networks
```
antmat: looks weird

insonmnia/worker/server.go
montekki: Description struct has Networks() and Network() methods.

insonmnia/worker/server.go
montekki: it does, but yet in the existing code https://github.com/sonm-io/core/blob/master/insonmnia/worker/server.go#L601 ```ssh.ParsePublicKey()``` is not checked for returned error, as you may see. In the event when taskinfo has no ssh key that would be the case and containerInfo.PublicKey whill be ```nil```. So when we restore the task the same logic has to be applied: if no key is in the storage, containerInfo.PublicKey is also ```nil```. This logic definitely has to be fixed, but I don't that this PR is the place to do it.

insonmnia/worker/server.go
montekki: It is a matter of taste: some ugly wrapping and unwrapping will be done in any case. But when adding more fields to container info that have to have this ugly marshalling, you wouldn't have to add more new wrapper types for those fields.

insonmnia/worker/server.go
montekki: That would be handy if it helped getting rid of this whole new ugly ```descriptionMarshallable``` type, but it looks that the ```Networks``` field remains another obstacle in the way here, so in either way the type stays until the ```Networks\Networks()``` disambiguation is resolved.

PR #1310, opened 2018-08-14
feat: close unprofitable deals in Connor

sshaman1101:
Chained with #1306 

This commit adds price checking for Connor's deals.
If mining profit becomes less that we paying for a deal -
Connor will close it without adding the supplier to the blacklist.


connor/engine.go
```diff
@@ -487,7 +489,25 @@ func (e *engine) checkDealStatus(ctx context.Context, log *zap.Logger, dealID *s
...
+	if dealStatus.GetDeal().GetStatus() == sonm.DealStatus_DEAL_ACCEPTED {
+		deal := e.dealFactory.FromDeal(dealStatus.GetDeal())
+		if deal.isReplaceable(e.priceProvider.GetPrice(), e.cfg.Market.PriceControl.DealCancelThreshold) {
+			log := log.Named("price-deviation")
+			log.Info("too much price deviation detected: closing deal",
+				zap.String("actual_price", e.priceProvider.GetPrice().String()),
+				zap.String("current_price", deal.restorePrice().String()))
+
+			if err := e.antiFraud.FinishDeal(deal.Unwrap()); err != nil {
+				log.Warn("failed to finish deal", zap.Error(err))
```
antmat: I think we need to handle this error properly

PR #1338, opened 2018-08-23
refactor: extract common parts from LogProcessor

sshaman1101:
Implementing `Processor` interface for another container forces to do a lot of shameless copy-paste.
Taking into consideration that log parsers one differs from another almost only by logline parsing code,
it will be better to extract reusable parts and be able to swap only the code that directly works with log lines.

As this commit aimed to deal with copy-paste, I'd also move log reader to the separated package.
I cannot put logReader into the `utils` neither `xgrpc`, what would look pretty logical.
But we'll face an issue with cycle imports in that case.

### DIFF
--- a/cmd/cli/commands/tasks.go
+++ b/cmd/cli/commands/tasks.go
@@ -263,27 +263,6 @@ func (m *logWriter) Write(p []byte) (n int, err error) {
 	return m.writer.Write(p)
 }
 
-type logReader struct {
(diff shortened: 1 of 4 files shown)


util/xlogs/reader.go
```diff
@@ -0,0 +1,33 @@
+package xlogs
```
3Hren: I'd rather move this helper to `proto` folder.

PR #1342, opened 2018-08-23
fix: allow NetworkIn only with public IPv4 addresses

oopcode:
This PR makes it possible for worker to sell AskPlans with `NetworkIn` only if there's at least one public IPv4 address.


3Hren:
No need for this for now, sorry.

PR #1352, opened 2018-08-27
fix: proper deal closing when price is changed

sshaman1101:
such problem occurs in the following conditions: Connor detects price deviation and starts replacing orders, also it starts to close non-profitalbe deals. When the deal is closed, a worker on another side can pick the order that we schedule for cancellation. This order is turned into a deal that immediately will be closed because of the low price. So this shitfall will perform until all of the orders isn't replaced. This commit fixes that behavior: now Connor checks that the cancel chan is empty before really closing an active deal.


antmat:
crutch, but ok for now. Consider thinking on a better solution

PR #1353, opened 2018-08-27
fix: remove repo & tag from uploaded image

oopcode:
Currently overseer allows users to push images with repo & tag info, which means that pushed images can overwrite existing containers with same repo & tag values (and that's a security breach). This PR removes repo & tag info from pushed images so that sha256 is always used as image identifier.

@antmat PTAL

### DIFF
--- a/insonmnia/worker/overseer.go
+++ b/insonmnia/worker/overseer.go
@@ -16,6 +16,7 @@ import (
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
 	"github.com/docker/go-connections/nat"
+	"github.com/docker/go-units"
 	"github.com/gliderlabs/ssh"
 	log "github.com/noxiouz/zapctx/ctxlog"
 	"github.com/pkg/errors"
@@ -661,7 +662,7 @@ func (m *prunedImage) Read(p []byte) (n int, err error) {
 
 func (m *prunedImage) load() (err error) {
 	if m.readingFile {
-		var in = make([]byte, 1<<15)
+		var in = make([]byte, units.KB*32)
 		n, err := m.image.Read(in)
 		if err == io.EOF {
 			// End of current file, switch to next header.
@@ -689,8 +690,8 @@ func (m *prunedImage) load() (err error) {
 	if err = m.writer.WriteHeader(hdr); err != nil {
 		return fmt.Errorf("failed to write header: %v", err)
 	}
-	// Other header types are not followed by data body.
-	if hdr.Typeflag == tar.TypeReg || hdr.Typeflag == tar.TypeRegA {
+	// Avoid reading zero length files.
+	if hdr.Size > 0 && (hdr.Typeflag == tar.TypeReg || hdr.Typeflag == tar.TypeRegA) {
 		m.readingFile = true
 	}
 
@@ -700,6 +701,9 @@ func (m *prunedImage) load() (err error) {
 // pruneManifest removes repository and tag data from image manifest.json to prevent
 // image overwriting.
 func (m *prunedImage) pruneManifest(hdr *tar.Header) error {
+	if hdr.Size > units.MB*100 {
+		return errors.New("manifest.json larger than 100MB, aborting")
+	}
 	var manifest = bytes.NewBuffer(nil)
 	if _, err := io.Copy(manifest, m.image); err != nil {
 		return err
--- a/insonmnia/worker/overseer.go
+++ b/insonmnia/worker/overseer.go
@@ -1,6 +1,8 @@
 package worker
 
 import (
+	"archive/tar"
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


insonmnia/worker/overseer.go
```diff
@@ -657,3 +663,111 @@ func (o *overseer) OnDealFinish(ctx context.Context, containerID string) error {
...
+			}
+			out = append(out, tarHeader2file{header: header, file: buf})
+		}
+	}
+
+	return nil, errors.New("unexpected exit")
+}
+
+func (o *overseer) compress(decompressed []tarHeader2file) (io.Reader, error) {
+	buf := bytes.NewBuffer([]byte{})
```
antmat: it's quite expensive in terms of memory usage and can fail with oom. Please consider buffering on disk.

insonmnia/worker/overseer.go
```diff
@@ -657,3 +663,109 @@ func (o *overseer) OnDealFinish(ctx context.Context, containerID string) error {
 func (o *overseer) Logs(ctx context.Context, id string, opts types.ContainerLogsOptions) (io.ReadCloser, error) {
 	return o.client.ContainerLogs(ctx, id, opts)
 }
+
+func (o *overseer) removeImageRepoAndTag(image io.Reader) (io.Reader, error) {
```
3Hren: This function should be out of the `overseer` scope because it doesn't use any of its fields nor methods.

insonmnia/worker/overseer.go
```diff
@@ -657,3 +663,109 @@ func (o *overseer) OnDealFinish(ctx context.Context, containerID string) error {
...
+
+	compressed, err := o.compress(pruned)
+	if err != nil {
+		return nil, fmt.Errorf("failed to re-compress image: %v", err)
+	}
+
+	return compressed, nil
+}
+
+func (o *overseer) extract(r io.Reader) ([]tarHeader2file, error) {
```
3Hren: This too.

insonmnia/worker/overseer.go
```diff
@@ -657,3 +663,109 @@ func (o *overseer) OnDealFinish(ctx context.Context, containerID string) error {
...
+			var buf = bytes.NewBuffer([]byte{})
+			if _, err := io.Copy(buf, tr); err != nil {
+				return nil, err
+			}
+			out = append(out, tarHeader2file{header: header, file: buf})
+		}
+	}
+}
+
+func (o *overseer) compress(decompressed []tarHeader2file) (io.Reader, error) {
```
3Hren: And this.

insonmnia/worker/overseer.go
```diff
@@ -657,3 +663,109 @@ func (o *overseer) OnDealFinish(ctx context.Context, containerID string) error {
...
+			if _, err := io.Copy(tw, h2f.file); err != nil {
+				return nil, fmt.Errorf("failed to add contents for header `%s`: %v", h2f.header.Name, err)
+			}
+		}
+	}
+
+	return buf, nil
+}
+
+func (o *overseer) getPrunedImageManifest(manifest []byte) ([]byte, error) {
```
3Hren: And this.

oopcode:
rewritten with streaming implementation

insonmnia/worker/overseer.go
```diff
@@ -657,3 +659,88 @@ func (o *overseer) OnDealFinish(ctx context.Context, containerID string) error {
...
+		}
+		return copy(p, m.buf.Bytes()), nil
+	} else {
+		// Try getting next header.
+		if m.currHeader, err = m.image.Next(); err == io.EOF {
+			// End of archive, write footer, exit with EOF.
+			if err = m.writer.Close(); err != nil {
+				return 0, fmt.Errorf("failed to write footer: %v", err)
+			}
+			return copy(p, m.buf.Bytes()), io.EOF
```
antmat: looks useless, because buf was reset

insonmnia/worker/overseer.go
```diff
@@ -657,3 +659,88 @@ func (o *overseer) OnDealFinish(ctx context.Context, containerID string) error {
...
+	if err != nil {
+		return 0, fmt.Errorf("failed to marshal manifest.json: %v", err)
+	}
+	m.currHeader.Size = int64(len(marshaled))
+	if err = m.writer.WriteHeader(m.currHeader); err != nil {
+		return 0, fmt.Errorf("failed to write header: %v", err)
+	}
+	n, err = m.writer.Write(marshaled)
+
+	return copy(p, m.buf.Bytes()), err
```
antmat: what if `len(p)` is less than m.buf.Bytes() ?

insonmnia/worker/overseer.go
```diff
@@ -657,3 +659,88 @@ func (o *overseer) OnDealFinish(ctx context.Context, containerID string) error {
...
+		if err != nil {
+			return 0, fmt.Errorf("failed to read header: %v", err)
+		}
+		if m.currHeader.Name == "manifest.json" {
+			return m.pruneManifest(p)
+		}
+		if err = m.writer.WriteHeader(m.currHeader); err != nil {
+			return 0, fmt.Errorf("failed to write header: %v", err)
+		}
+		return copy(p, m.buf.Bytes()), nil
```
antmat: same here, if `len(p)` is less than `len(m.buf.Bytes)` you will loose data.

insonmnia/worker/overseer.go
oopcode: Close() flushes, writes padding and footer, buf is not empty after this call.

PR #1354, opened 2018-08-27
feat: parallel order and deal purging

antmat:
This PR extends order and deal api, providing ability to remove several orders and deals at a time (including full purge). This is done on node in parallel manner reducing time needed for operation.


util/multierror/error.go
```diff
@@ -32,3 +37,14 @@ func errorFormat(errs []error) string {
 	return strings.Join(s, ", ")
 
 }
+
+type TSMultiError struct {
+	*multierror.Error
+	mu sync.Mutex
```
3Hren: The mutex is placed before the variable it protects.

util/multierror/error.go
```diff
@@ -32,3 +37,14 @@ func errorFormat(errs []error) string {
...
+}
+
+func (m *TSMultiError) Append(errs ...error) {
```
3Hren: Non-thread-safe version returns itself, this - is not. Why introduce such inconsistency?

cmd/cli/commands/deals.go
```diff
@@ -234,40 +235,78 @@ var dealQuickBuyCmd = &cobra.Command{
 	},
 }
 
+func getBlacklistType() (pb.BlacklistType, error) {
+	switch blacklistTypeStr {
+	case "none":
```
3Hren: Why `none`, not `nobody` as it described in our proto?

cmd/cli/commands/deals.go
```diff
@@ -234,40 +235,78 @@ var dealQuickBuyCmd = &cobra.Command{
 	},
 }
 
+func getBlacklistType() (pb.BlacklistType, error) {
```
3Hren: Rewrite using the following algorithm:
1. Uppercase.
2. Prepend `BLACKLIST_`.
3. Use either `proto.UnmarshalJSONEnum` or `BlacklistType_value`.

Then it will be more flexible to future changes.

cmd/cli/commands/orders.go
```diff
@@ -119,10 +120,19 @@ var orderCancelCmd = &cobra.Command{
...
 		}
 
-		orderID := args[0]
-		_, err = market.CancelOrder(ctx, &pb.ID{Id: orderID})
-		if err != nil {
-			return fmt.Errorf("cannot cancel order on Marketplace: %v", err)
+		request := &pb.OrderIDs{
+			Ids: make([]*pb.BigInt, 0, len(args)),
+		}
+		for _, idStr := range args {
```
3Hren: Great idea - decompose this. Then this will be oneliner. Also, there are several places where such boilerplate exists.

insonmnia/node/deals.go
```diff
@@ -88,6 +91,66 @@ func (d *dealsAPI) Finish(ctx context.Context, req *pb.DealFinishRequest) (*pb.E
 	return &pb.Empty{}, nil
 }
 
+func (d *dealsAPI) FinishDeals(ctx context.Context, req *pb.DealsFinishRequest) (*pb.Empty, error) {
+	return d.finishDeals(ctx, req.GetDealInfo())
+}
+
+func (d *dealsAPI) finishDeals(ctx context.Context, deals []*pb.DealFinishRequest) (*pb.Empty, error) {
+	d.log.Debugf("finishing %d deals", len(deals))
```
3Hren: Why? There is request logger, where more detailed info shown.

insonmnia/node/deals.go
```diff
@@ -88,6 +91,66 @@ func (d *dealsAPI) Finish(ctx context.Context, req *pb.DealFinishRequest) (*pb.E
...
+					merr.Append(errors.New("zero or nil deal id specified"))
+					continue
+				}
+				idStr := info.Id.Unwrap().String()
+				d.log.Debugw("finishing deal", zap.String("id", idStr))
+				if err := d.remotes.eth.Market().CloseDeal(ctx, d.remotes.key, info.GetId().Unwrap(), info.GetBlacklistType()); err != nil {
+					merr.Append(fmt.Errorf("cannot cancel order with id %s: %v", idStr, err))
+				}
+			}
+			wg.Done()
```
3Hren: `defer wg.Done` at the beginning.

insonmnia/node/deals.go
```diff
@@ -88,6 +91,66 @@ func (d *dealsAPI) Finish(ctx context.Context, req *pb.DealFinishRequest) (*pb.E
...
+		}()
+	}
+	for _, info := range deals {
+		ch <- info
+	}
+	close(ch)
+	wg.Wait()
+	if len(merr.WrappedErrors()) > 0 {
```
3Hren: So why do we need this check where the name of the `ErrorOrNil ` method itself says that it performs emptiness check itself?

insonmnia/node/deals.go
```diff
@@ -88,6 +91,66 @@ func (d *dealsAPI) Finish(ctx context.Context, req *pb.DealFinishRequest) (*pb.E
...
+	}
+
+	return &pb.Empty{}, nil
+}
+
+func (d *dealsAPI) PurgeDeals(ctx context.Context, req *pb.DealsPurgeRequest) (*pb.Empty, error) {
+	deals, err := d.remotes.dwh.GetDeals(ctx, &pb.DealsRequest{
+		Status:     pb.DealStatus_DEAL_ACCEPTED,
+		ConsumerID: pb.NewEthAddress(crypto.PubkeyToAddress(d.remotes.key.PublicKey)),
+		Limit:      dwh.MaxLimit,
```
3Hren: This will not always work as you expect. You should implement DWH cursor for such case (see Optimus).

proto/node.proto
```diff
@@ -51,7 +51,12 @@ service DealManagement {
     // Status produces a detailed info about deal with given ID.
     rpc Status(BigInt) returns (DealInfoReply) {}
     // Finish finishes a deal with given ID
+    // Deprecated: use FinishDeals instead
```
3Hren: I don't think we should deprecate non-pluralized method.

util/multierror/error.go
antmat: because this one is a mutating method, not a standalone function. Returning itself would be misleding, because usage is different(you don't need to pss itself as a first param).

cmd/cli/commands/deals.go
antmat: this was leaved as it was previous in the code - just reusble method was extracted

cmd/cli/commands/deals.go
antmat: you know, it would be out of the scope, because code was just extracted unchanged, but ok, i'll do it

insonmnia/node/deals.go
antmat: just to return nil *pb.Empty{} in case of error. I don' like this too, but (value,nil)/(nil, error) is amore common pattern than (value, nil)/(value, error)

proto/node.proto
antmat: it's functionality is dublicated, plural methods are quite easy to use and do the same thing if you pass single value. I still think we should deprecate them.

cmd/cli/commands/orders.go
antmat: not really oneliner, because go has no `map`

PR #1360, opened 2018-08-29
fix: allow grpc logger to truncate specific methods output

oopcode:
Short method names should be provided, i.e., `PushTask`, not `/sonm.Worker/PushTask`.


util/xgrpc/options.go
```diff
@@ -240,40 +240,48 @@ func verifyUnaryInterceptor() grpc.UnaryServerInterceptor {
 //
 // Note, that to enable tracing logging you should specify this option AFTER
 // trace interceptors.
-func RequestLogInterceptor(log *zap.Logger) ServerOption {
+func RequestLogInterceptor(log *zap.Logger, truncatedMethods []string) ServerOption {
+	var truncatedMethodsSet = map[string]bool{}
```
antmat: It is more common to use `map[string]struct{}` for sets (yep, go sucks)

util/xgrpc/options.go
```diff
@@ -240,40 +240,48 @@ func verifyUnaryInterceptor() grpc.UnaryServerInterceptor {
...
+		o.interceptors.u = append(o.interceptors.u, requestLogUnaryInterceptor(log.Sugar(), truncatedMethodsSet))
+		o.interceptors.s = append(o.interceptors.s, requestLogStreamInterceptor(log.Sugar(), truncatedMethodsSet))
 	}
 }
 
-func requestLogUnaryInterceptor(log *zap.SugaredLogger) grpc.UnaryServerInterceptor {
+func requestLogUnaryInterceptor(log *zap.SugaredLogger, truncatedMethods map[string]bool) grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
-		executeRequestLogging(ctx, req, info.FullMethod, log)
+		executeRequestLogging(ctx, req, info.FullMethod, log, truncatedMethods[MethodInfo(info.FullMethod).Method])
```
antmat: did you test it? Because go panics on missing key

oopcode: nuh-uh no it doesn't

antmat: yep, it looks like I was not drunk

PR #1367, opened 2018-08-30
chore: worker's code cleanup

sshaman1101:
- Removed several stale TODOs, which was introduced
during previous large refactorings.
- Method's docstrings fixed to match the golang's standards.
- Import and variable names clashing avoided.

### DIFF
--- a/insonmnia/worker/overseer.go
+++ b/insonmnia/worker/overseer.go
@@ -154,9 +154,9 @@ type ContainerMetrics struct {
 }
 
 func (m *ContainerMetrics) Marshal() *pb.ResourceUsage {
-	network := make(map[string]*pb.NetworkUsage)
+	networkUsage := make(map[string]*pb.NetworkUsage)
 	for i, n := range m.net {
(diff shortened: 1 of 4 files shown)


insonmnia/worker/overseer.go
```diff
@@ -154,9 +154,9 @@ type ContainerMetrics struct {
 }
 
 func (m *ContainerMetrics) Marshal() *pb.ResourceUsage {
-	network := make(map[string]*pb.NetworkUsage)
+	net := make(map[string]*pb.NetworkUsage)
```
oopcode: this will clash with std `net` if somebody decides to use it here

insonmnia/worker/server.go
```diff
@@ -972,6 +969,9 @@ func (m *Worker) setupRunningContainers() error {
 	defer dockerClient.Close()
 
 	containers, err := dockerClient.ContainerList(m.ctx, types.ContainerListOptions{})
+	if err != nil {
```
oopcode: wow

PR #1368, opened 2018-08-30
fix: show allocated resources in TaskStatus

sshaman1101:
This commit fills the `AllocatedResources` field from the deal's ask-plan.
Actual resources info gives an ability to show what part of deal's resources are acquired by task, at least what GPUs is used.

### DIFF
--- a/insonmnia/resource/pool.go
+++ b/insonmnia/resource/pool.go
@@ -14,12 +14,14 @@ import (
 )
 
 type Scheduler struct {
-	OS            *hardware.Hardware
-	mu            sync.Mutex
-	pool          *pool
+	OS   *hardware.Hardware
+	mu   sync.Mutex
+	pool *pool
+	// taskToAskPlan maps task ID to ask plan ID
 	taskToAskPlan map[string]string
-	askPlanPools  map[string]*pool
-	log           *zap.SugaredLogger
+	// askPlanPools maps ask plan' ID to allocated resource pool
+	askPlanPools map[string]*pool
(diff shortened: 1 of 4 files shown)


insonmnia/resource/pool.go
```diff
@@ -163,6 +163,17 @@ func (m *Scheduler) ReleaseTask(taskID string) error {
 	return nil
 }
 
+func (m *Scheduler) ResourceByTask(id string) (*sonm.AskPlanResources, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	res, ok := m.pool.used[id]
+	if !ok {
+		return nil, fmt.Errorf("failed to get resources")
```
antmat: `return nil, fmt.Errorf("failed to get resources for task %s: no such task", id)`

insonmnia/resource/pool.go
```diff
@@ -163,6 +163,17 @@ func (m *Scheduler) ReleaseTask(taskID string) error {
 	return nil
 }
 
+func (m *Scheduler) ResourceByTask(id string) (*sonm.AskPlanResources, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	res, ok := m.pool.used[id]
```
antmat: this one is global pool for ask plans. You need to do smth like
```
ask, ok := m.taskToAskPlan[id]
//check ok
pool, ok := m.askPlanPools[ask.GetId()]
//check ok
res, ok := pool.used[id]
```

PR #1369, opened 2018-08-30
aux: extend ethclient GetTransactionReceipt method

sokel:
Extend CustomEthereumClient GetTransactionReceipt method with new fields.
added fields:
 * BlockHash
 * From
 * To
 * TransactionIndex


sokel:
I use these fields with client at blockchain explorer.

PR #1380, opened 2018-08-31
chore: update default task.yaml

3Hren:


task.yaml
```diff
@@ -1,44 +1,76 @@
...
   # Env variables that will be passed to container on start.
+  # Optional.
   env:
     param1: value1
     param2: value2
-    param3: value3
-  # Add ability to pull container back.
+  # Adds an ability to pull container back.
+  # Optional.
   commit_on_stop: true
```
sshaman1101: IMO nobody really using it, maybe disable it by default?

PR #1389, opened 2018-09-01
aux: debug whitelist for Connor

sshaman1101:
USE ONLY FOR DEBUGGING PURPOSES AND NEVER USE IN PRODUCTION.

Connor debugging will be pretty easy if deals with test suppliers in the Testent will not be closed (because we should remove it from the blacklist, restart Connor, restart tasks, etc..
and it does the whole development process really sucks).

### DIFF
--- a/connor/antifraud/antifraud.go
+++ b/connor/antifraud/antifraud.go
@@ -131,10 +131,17 @@ func (m *antiFraud) checkDeals(ctx context.Context) error {
 			log.Debug("task quality is less that required: detected by pool reports")
 		}
 
+		supplier := dealMeta.deal.GetSupplierID().Unwrap()
 		logQualityMetrics := []zapcore.Field{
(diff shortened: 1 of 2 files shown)


connor/antifraud/antifraud.go
```diff
@@ -207,7 +207,19 @@ func (m *antiFraud) DealOpened(deal *sonm.Deal) error {
 	return nil
 }
 
-func (m *antiFraud) FinishDeal(ctx context.Context, deal *sonm.Deal, optsFlags flags) error {
+func (m *antiFraud) FinishDeal(ctx context.Context, deal *sonm.Deal, crutch func(), optsFlags flags) error {
+	if !optsFlags.SkipCrutch() {
```
oopcode: SkipCrutch? Seriously? : )

PR #1441, opened 2018-09-07
fix: keep container info after task stop

oopcode:
`TaskStop` should not make worker forget about the task (i.e., worker should know about stopped tasks after restart if deal is still active).

### DIFF
--- a/insonmnia/worker/server.go
+++ b/insonmnia/worker/server.go
@@ -801,11 +801,6 @@ func (m *Worker) StopTask(ctx context.Context, request *pb.ID) (*pb.Empty, error
 	}
 
 	m.setStatus(&pb.TaskStatusReply{Status: pb.TaskStatusReply_FINISHED}, request.Id)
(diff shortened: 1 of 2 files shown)


insonmnia/worker/server.go
```diff
@@ -801,11 +801,6 @@ func (m *Worker) StopTask(ctx context.Context, request *pb.ID) (*pb.Empty, error
 	}
 
 	m.setStatus(&pb.TaskStatusReply{Status: pb.TaskStatusReply_FINISHED}, request.Id)
-	_, err := m.storage.Remove(containerInfo.TaskId)
```
antmat: so, it will be stored forever?

oopcode: containers will be removed when the deal is [finished](https://github.com/sonm-io/core/blob/fix/keep-task-info-after-stop/insonmnia/worker/server.go#L410)

PR #1512, opened 2018-09-18
aux: adapt Connor for new benchmark

sshaman1101:


3Hren:
Is this safe enough to be merged?

PR #1516, opened 2018-09-18
aux: proper systemd config

abefimov:


sshaman1101:
Works with SONM Installer and Ubuntu 18.04 (systemd 229).

PR #1525, opened 2018-09-19
feat: gRPC rate limiter

### LINKED ISSUE #1495: Rate limit gRPC
DEV-819

3Hren:
This commit allows restricting Worker and Node services usage by specifying rate limiter interceptor, which can be configured for taking in action both global and precise settings - for each method differently.
This is required to avoid some dumb DOS attempts, however, not ideal.
The idea is to keep RPS counter for each remote peer that is identified by ETH address, which we can extract from the request. The counter has EWMA internally, which is good for an unpredictable load profile. When this EWMA exceeds the specified threshold an "Unavailable" error is returned.

Closes #1495.


noxiouz:
Is InTap handler friendly enough? To save resources it's better:https://godoc.org/google.golang.org/grpc#InTapHandle

3Hren:
Yes, I think it will be better. The only suspicious thing about that link it that everything is marked as EXPERIMENTAL.

PR #1540, opened 2018-09-21
feat: dwh stats handle

oopcode:
The new `GetStats()` handle returns current DWH stats.


insonmnia/dwh/sql.go
```diff
@@ -1134,6 +1134,61 @@ func (m *sqlStorage) GetLastEvent(conn queryConn) (*blockchain.Event, error) {
 	}, nil
 }
 
+func (m *sqlStorage) getStats(conn queryConn) (*sonm.DWHStatsReply, error) {
+	var (
+		numCurrDealsQ, _, _ = m.builder().Select("count(Id)").From("Deals").Where("Status=1").Prefix("(").Suffix(")").ToSql()
```
sshaman1101: maybe pass const with status via `.Where`'s  args?

oopcode: not sure how this is any better, do you have any ideas?

oopcode: ok, sounds reasonable

insonmnia/dwh/sql.go
```diff
@@ -1134,6 +1134,61 @@ func (m *sqlStorage) GetLastEvent(conn queryConn) (*blockchain.Event, error) {
...
+		numDealsQ, _, _     = m.builder().Select("count(Id)").From("Deals").Prefix("(").Suffix(")").ToSql()
```
avsigaev: add filter: 'where status = 1'.

We need number of current active deals here

insonmnia/dwh/server.go
```diff
@@ -399,3 +425,10 @@ func (m *DWH) GetWorkers(ctx context.Context, request *sonm.WorkersRequest) (*so
 
 	return &sonm.WorkersReply{Workers: workers, Count: count}, nil
 }
+
+func (m *DWH) GetStats(ctx context.Context, request *sonm.Empty) (*sonm.DWHStatsReply, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	return m.stats, nil
```
antmat: this is returned by pointer and can be serialized during modification

PR #1555, opened 2018-09-26
patch: tasks GC

oopcode:
This PR introduces a GC that kills and removes containers that might, for some reason, be running after the deal was closed.

### DIFF
--- a/insonmnia/worker/overseer.go
+++ b/insonmnia/worker/overseer.go
@@ -217,6 +217,10 @@ type Overseer interface {
 	// Stop terminates the container.
 	Stop(ctx context.Context, containerID string) error
 
+	// PurgeContainer makes Overseer completely remove and forget the container, no matter what
+	// was known about that container.
+	PurgeContainer(ctx context.Context, containerID string) error
+
 	// OnDealFinish makes all cleanup related to closed deal
 	OnDealFinish(ctx context.Context, containerID string) error
 
@@ -601,6 +605,28 @@ func (o *overseer) Stop(ctx context.Context, containerid string) error {
 	}
 
 	return descriptor.Kill(ctx)
+}
+
+func (o *overseer) PurgeContainer(ctx context.Context, containerID string) error {
+	var result = multierror.NewMultiError()
+
+	o.mu.Lock()
+	// Unconditionally remove all (possibly left) info about the container.
+	delete(o.containers, containerID)
+	if status, ok := o.statuses[containerID]; ok {
+		close(status)
+	}
+	delete(o.statuses, containerID)
+	o.mu.Unlock()
+
+	if err := o.client.ContainerKill(ctx, containerID, "SIGKILL"); err != nil {
+		result = multierror.Append(result, fmt.Errorf("failed to kill container: %v", err))
+	}
+	if err := o.client.ContainerRemove(ctx, containerID, types.ContainerRemoveOptions{}); err != nil {
+		result = multierror.Append(result, fmt.Errorf("failed to remove container: %v", err))
+	}
+
+	return result.ErrorOrNil()
 }
 
 func (o *overseer) OnDealFinish(ctx context.Context, containerID string) error {
--- a/insonmnia/worker/overseer.go
+++ b/insonmnia/worker/overseer.go
@@ -615,8 +615,8 @@ func (o *overseer) PurgeContainer(ctx context.Context, containerID string) error
 	delete(o.containers, containerID)
 	if status, ok := o.statuses[containerID]; ok {
 		close(status)
+		delete(o.statuses, containerID)
 	}
-	delete(o.statuses, containerID)
 	o.mu.Unlock()
 
 	if err := o.client.ContainerKill(ctx, containerID, "SIGKILL"); err != nil {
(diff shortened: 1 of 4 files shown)


insonmnia/worker/server.go
```diff
@@ -1096,6 +1096,69 @@ func (m *Worker) setupRunningContainers() error {
...
+			if !ok {
+				log.G(m.ctx).Warn("container has corrupted deal id",
+					zap.String("container_id", container.ID), zap.Any("deal_id", dealID))
+				continue
+			}
+
+			dealInfo, err := m.getDealInfo(sonm.NewBigInt(bigDealID))
+			if err != nil {
+				log.G(m.ctx).Warn("failed to get deal info", zap.String("container_id", container.ID),
+					zap.Any("deal_id", dealID))
```
sshaman1101: also log `err`

insonmnia/worker/server.go
```diff
@@ -1096,6 +1096,69 @@ func (m *Worker) setupRunningContainers() error {
...
+				continue
+			}
+
+			if dealInfo.Deal.Status == sonm.DealStatus_DEAL_CLOSED {
+				log.G(m.ctx).Info("found running container assigned to closed deal, going to purge it",
+					zap.String("container_id", container.ID), zap.String("deal_id", dealID))
+				if err := m.ovs.PurgeContainer(m.ctx, container.ID); err != nil {
+					log.G(m.ctx).Warn("failed to purge container", zap.String("container_id", container.ID))
```
sshaman1101: log `err`

insonmnia/worker/server.go
```diff
@@ -1096,6 +1096,70 @@ func (m *Worker) setupRunningContainers() error {
...
+
+	for _, container := range containers {
+		if _, ok := container.Labels[overseerTag]; ok {
+			dealID, ok := container.Labels[dealIDTag]
+			if !ok {
+				log.G(m.ctx).Warn("container has no deal id info", zap.String("container_id", container.ID))
+				continue
+			}
+
+			bigDealID, ok := big.NewInt(0).SetString(dealID, 10)
```
antmat: sonm.NewBigIntFromString

insonmnia/worker/overseer.go
```diff
@@ -603,6 +607,28 @@ func (o *overseer) Stop(ctx context.Context, containerid string) error {
...
+func (o *overseer) PurgeContainer(ctx context.Context, containerID string) error {
+	var result = multierror.NewMultiError()
+
+	o.mu.Lock()
+	// Unconditionally remove all (possibly left) info about the container.
+	delete(o.containers, containerID)
+	if status, ok := o.statuses[containerID]; ok {
+		close(status)
+	}
+	delete(o.statuses, containerID)
```
antmat: move delete inside if

insonmnia/worker/server.go
```diff
@@ -1096,6 +1096,70 @@ func (m *Worker) setupRunningContainers() error {
 	return nil
 }
 
+func (m *Worker) tasksGC() error {
```
antmat: I think maybe it's better to do it only once on start

insonmnia/worker/overseer.go
```diff
@@ -603,6 +607,28 @@ func (o *overseer) Stop(ctx context.Context, containerid string) error {
 	return descriptor.Kill(ctx)
 }
 
+func (o *overseer) PurgeContainer(ctx context.Context, containerID string) error {
```
antmat: what's the difference of this function and OnDealFinish?

insonmnia/worker/server.go
```diff
@@ -1096,6 +1096,70 @@ func (m *Worker) setupRunningContainers() error {
...
+			}
+
+			bigDealID, ok := big.NewInt(0).SetString(dealID, 10)
+			if !ok {
+				log.G(m.ctx).Warn("container has corrupted deal id",
+					zap.String("container_id", container.ID), zap.Any("deal_id", dealID))
+				continue
+			}
+
+			dealInfo, err := m.getDealInfo(sonm.NewBigInt(bigDealID))
```
antmat: why don't you check if worker know about this deal instead of going to blockchain?
Also in your case you need to check if this is our deal

insonmnia/worker/overseer.go
oopcode: `OnDealFinish` assumes that worker has all possible info about the running container and will abort if some of this info is missing (which is not what we want).

insonmnia/worker/server.go
oopcode: wtf, `getDealInfo()` does not go to blockchain

insonmnia/worker/server.go
oopcode: we are trying to fix a state that results from race condition (that does not necessarily occur only when someone shuts the worker down)

insonmnia/worker/server.go
```diff
@@ -1096,6 +1096,70 @@ func (m *Worker) setupRunningContainers() error {
...
+}
+
+func (m *Worker) removeExpiredContainers() error {
+	dockerClient, err := client.NewEnvClient()
+	if err != nil {
+		return err
+	}
+	defer dockerClient.Close()
+
+	// Collect only running containers.
```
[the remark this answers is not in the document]
oopcode: because that was the task definition. if we try to gcollect all types of leftover containers, the implementation is going to be much more complicated.

insonmnia/worker/server.go
oopcode: no

PR #1578, opened 2018-10-02
aux: keep grpccmd staff closer to core

sshaman1101:
This commit adds `xcode` package which is used with generated code for the `autocli` tool.

Internals:
 * You must update protoc compiler plugin with the following command `go get -u github.com/sshaman1101/grpccmd/cmd/protoc-gen-grpccmd`.


3Hren:
Currently, the problem still remains: generated files in `sonm` package (aka `proto`) depend on `xcode`, which in its turn depends on many other things.

PR #1586, opened 2018-10-03
feat: worker stats collection

oopcode:
This PR allows for worker stats collection from specified addresses.


3Hren:
Rather than creating another GitHub dependency, I'd put this into a smart contract.

3Hren:
Seems like we've ended with pushing devices info into sidechain. Feel free to reopen if I'm wrong.

PR #1601, opened 2018-10-06
feat: show trace info in Rendezvous logs

3Hren:


insonmnia/logging/trace.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func WithTrace(ctx context.Context, log *zap.Logger) *zap.Logger {
+	span := opentracing.SpanFromContext(ctx)
+	if span == nil {
+		return log
+	}
+
+	spanContext, ok := span.Context().(basictracer.SpanContext)
+	if ok && spanContext.Sampled {
```
antmat: I doubt that check for Sampled is required here. Did you check if all records in this manner do have corresponding tracing fields?

PR #1604, opened 2018-10-08
feat: dwh out of sync mode

oopcode:
This PR makes DWH refuse to respond to user requests if the DB is out of sync with blockchain.


insonmnia/dwh/server.go
```diff
@@ -6,6 +6,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"github.com/sonm-io/core/proto"
```
sshaman1101: WHY U NOT SORT IMPORTS?

insonmnia/dwh/server.go
```diff
@@ -226,12 +228,47 @@ func (m *DWH) monitorStatistics() error {
 	}
 }
 
+func (m *DWH) monitorSync() error {
+	const maxBacklog = 50
+	tk := util.NewImmediateTicker(time.Minute)
```
sshaman1101: `defer tk.Stop()`

PR #1606, opened 2018-10-08
fix: don't init Connor's backends twice

sshaman1101:


3Hren:
Mixing between configuration and the result of such config. Can you just separate obtaining backends from the config itself?

PR #1621, opened 2018-10-10
feat: return REST errors as JSON

3Hren:
This commit slighty changes the REST server behavior by forcing to return errors as JSON for the sake of consistency.
Previously all errors were returned as a plain text. Now set set the application/json header and wrap the error into a JSON object.
Also the encoding/decoding logic was slightly changed - now if a Node is configured into secure mode it will return all responses as encoded byte array. This was done for symmetry, because one can have a client with encoder, and it expects the encoded response. Previously if a request failed in URI parsing stage a plain-text reply was returned. That was unexpected and completely helpless for people that try to work with secure Node in insecure manner. The error says something, but anyway you're unable to do anything with the Node unless you have a client-side encoder/decoder.

### DIFF
--- a/util/rest/server.go
+++ b/util/rest/server.go
@@ -135,19 +135,20 @@ func (s *Server) ServeHTTP(rw http.ResponseWriter, request *http.Request) {
 
 	switch response := response.(type) {
 	case error:
-		if code/100 == 4 || code/100 == 5 {
-			s.log.Warn(response.Error())
+		data, err := json.Marshal(map[string]string{
+			"error": response.Error(),
+		})
+		if err != nil {
+			panic("unreachable")
 		}
 
-		rw.Write([]byte(fmt.Sprintf(`{"error": "%s"}`, response.Error())))
+		rw.Write(data)
 	case []byte:
 		rw.Write(response)
 	default:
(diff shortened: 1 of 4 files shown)


util/rest/server.go
```diff
@@ -111,65 +117,64 @@ func (s *Server) RegisterService(interfacePtr, concretePtr interface{}) error {
...
+	rw.Header().Set("content-type", "application/json")
+	rw.WriteHeader(code)
+
+	switch response := response.(type) {
+	case error:
+		if code/100 == 4 || code/100 == 5 {
+			s.log.Warn(response.Error())
 		}
+
+		rw.Write([]byte(fmt.Sprintf(`{"error": "%s"}`, response.Error())))
```
antmat: use json marshaler

util/rest/server.go
```diff
@@ -111,65 +117,64 @@ func (s *Server) RegisterService(interfacePtr, concretePtr interface{}) error {
...
-			rw.Write([]byte("invalid uri provided"))
-			return
+
+	code, response := s.serveHTTP(request)
+	rw.Header().Set("content-type", "application/json")
+	rw.WriteHeader(code)
+
+	switch response := response.(type) {
+	case error:
+		if code/100 == 4 || code/100 == 5 {
```
antmat: seems useless

PR #1628, opened 2018-10-10
fix: no SIGSEGV while logging requests

3Hren:


antmat:
this should no be required with #1629 but whatever

PR #1646, opened 2018-10-17
aux: always restart services via systemd

sshaman1101:


3Hren:
I don't think we need this after #1647 be merged.

PR #1649, opened 2018-10-17
feat: add BID tag to deals list

### LINKED ISSUE #1493: Deal list - need output tag from bid order
1. Update DWH/deals table structure - add tag from BID order.
2. Change Node/CLI output for 'deal list' and 'deal status' (both simple and json format).

oopcode:
This PR adds `BidTag` to DWHDeal and makes this tag printable in `deals list`.

Closes #1493


3Hren:
Why it should be in DWHDeal?
Why there is no ask tags?

3Hren:
Actually, I think we don't need this because anyway there is a new `GetOrders` request in the DWH, which we can utilize.

PR #1681, opened 2018-11-01
feat: SSH keys management

3Hren:
This PR allows to specify untrusted and leaked ETH addresses that are not allowed to be used for remote SSH login. This includes our public test key, that is widely used for testing. More can be specified in Worker's config.

### DIFF
--- a/insonmnia/worker/options.go
+++ b/insonmnia/worker/options.go
@@ -35,6 +35,10 @@ const (
 	exportKeystorePath    = "/var/lib/sonm/worker_keystore"
 )
 
+var (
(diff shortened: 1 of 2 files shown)


insonmnia/worker/options.go
```diff
@@ -35,6 +35,10 @@ const (
 	exportKeystorePath    = "/var/lib/sonm/worker_keystore"
 )
 
+var (
+	leakedInsecureKey = common.HexToAddress("0x8125721c2413d99a33e351e1f6bb4e56b6b633fd")
```
sshaman1101: maybe will be better to put it into the default worker's config rather than hardcoding the address?

PR #1697, opened 2018-11-12
feat: ability to collect metrics from DRI devices

sshaman1101:
Now we're trying to find hwmon path on the device initialization.
If we can do this, the monitoring interface for this card became available.

The `Monitoring` method reads some data from the `/sys/dev` internals, all behavior has been ported from the [ROC-smi tool](https://github.com/RadeonOpenCompute/ROC-smi).

### DIFF
--- a/insonmnia/worker/gpu/dri.go
+++ b/insonmnia/worker/gpu/dri.go
@@ -64,10 +64,7 @@ func NewDRICard(num int, name, path string) (DRICard, error) {
 		return DRICard{}, err
 	}
 
-	err = c.collectHwmonPath()
-	if err != nil {
-		return DRICard{}, err
-	}
+	c.collectHwmonPath()
 
 	return c, nil
 }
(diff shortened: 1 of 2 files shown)


insonmnia/worker/gpu/dri.go
```diff
@@ -137,6 +148,88 @@ func (card *DRICard) collectPCIBusID() error {
...
+
+	dir, err := ioutil.ReadDir(p)
+	if err != nil {
+		// if we can't find such directory - just skip it,
+		// hwmon dir may be not present for on-board card.
+		return nil
+	}
+
+	for _, fd := range dir {
+		if fd.IsDir() && strings.Contains(fd.Name(), "hwmon") {
```
3Hren:
```suggestion
		if fd.IsDir() && strings.HasPrefix(fd.Name(), "hwmon") {
```

insonmnia/worker/gpu/dri.go
```diff
@@ -137,6 +148,88 @@ func (card *DRICard) collectPCIBusID() error {
 	return errors.New("cannot find PCI slot name into the file")
 }
 
+func (card *DRICard) collectHwmonPath() error {
```
3Hren: This function actually does not return any error.

PR #1698, opened 2018-11-13
fix: properly restore worker state on restart

antmat:
This PR fixes the bug when worker forgets about his tasks on restart

### DIFF
--- a/insonmnia/worker/salesman/salesman.go
+++ b/insonmnia/worker/salesman/salesman.go
@@ -371,7 +371,7 @@ func (m *Salesman) restoreState(ctx context.Context) error {
 
 	for _, plan := range m.askPlans {
 		// We reset deal id here in order to restart order processing and reassign deal properly to ask plan.
-		// we could do it here synchronously, but this would slower startup.
+		// We could do it here synchronously, but this would slower startup.
 		plan.DealID = nil
 		if err := m.resources.Consume(plan); err != nil {
 			m.log.Warnf("dropping ask plan %s due to resource changes", plan.ID)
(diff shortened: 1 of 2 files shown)


insonmnia/worker/salesman/salesman.go
```diff
@@ -370,6 +370,9 @@ func (m *Salesman) restoreState(ctx context.Context) error {
 	m.log.Infow("removed no longer used networks", zap.Any("networks", *pruneReply))
 
 	for _, plan := range m.askPlans {
+		// We reset deal id here in order to restart order processing and reassign deal properly to ask plan.
+		// we could do it here synchronously, but this would slower startup.
```
3Hren:
```suggestion
		// We could do it here synchronously, but this would slower startup.
```

insonmnia/worker/server.go
```diff
@@ -183,12 +183,15 @@ func NewWorker(opts ...Option) (m *Worker, err error) {
...
 
-	if err := m.setupRunningContainers(); err != nil {
+	// First we setup salesman here to restore all known ask plans to perform container restoration.
+	if err := m.setupSalesman(); err != nil {
 		m.Close()
 		return nil, err
 	}
 
-	if err := m.setupSalesman(); err != nil {
+	// at this step all ask plans should be restored and task resources can be successfully consumed by scheduler.
```
```suggestion
	// At this step, all ask plans should be restored and task resources can be successfully consumed by the scheduler.
```

insonmnia/worker/server.go
```diff
@@ -183,12 +183,15 @@ func NewWorker(opts ...Option) (m *Worker, err error) {
...
+	// Note: if some ask plans were removed due to resource lack corresponding containers will be removed in setupRunningContainers
```
```suggestion
	// Note: if some ask plans were removed due to resource lack corresponding containers will be removed in "setupRunningContainers".
```

PR #1707, opened 2018-11-14
feat: allow to disable Worker benchmarking in dev mode

3Hren:


3Hren:
I'm a bit frustrated while developing the Worker in a REPL way.

PR #1715, opened 2018-11-15
feat: background metrics collector for worker

sshaman1101:
This PR Introduces hardware metrics of the worker, consists of the following parts:
* MetricsHandler plugin bound to worker's instance, it's responsible for handling hardware-specific metrics providers (currently for GPUs only), updating the whole set of metrics by the timer, return last known state.
* gRPC method for WorkerManagement API, can be used to retrieve metrics by the master or admin address.

Things to do in the next PRs:
* Improve the quality of metrics.
* Make pretty-printer for `sonmcli worker metrics` handle.
* Extend ACLs, provide an ability to connect to workers from the metrics collector.


cmd/cli/commands/worker_metrics.go
```diff
@@ -0,0 +1,28 @@
...
+		if err != nil {
+			return fmt.Errorf("cannot get metrics: %v", err)
+		}
+
+		data, err := yaml.Marshal(metrics)
+		if err != nil {
+			return fmt.Errorf("failed to marshal metrics into YAML: %v", err)
+		}
+
+		cmd.Println(string(data))
```
antmat: what about json?

proto/worker.proto
```diff
@@ -202,3 +204,43 @@ message DebugStateReply {
...
+    uint64 total = 1;
+    uint64 free = 2;
+}
+
+message DiskMetrics {
+    uint64 total = 1;
+    uint64 free = 2;
+}
+
+message WorkerMetricsResponse{
```
antmat: do we really need typed response here? Maybe somethig like {name, value, aggregation type}?

insonmnia/worker/metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+			}
+		}
+	}()
+}
+
+func (m *Handler) update(ctx context.Context) {
+	m.logger.Debug("updating hardware metrics")
+	isError := false
+
+	g, err := m.updateGPUMetrics()
```
antmat: no context here?

insonmnia/worker/metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+		isError = true
+	}
+
+	r, err := m.updateRAMMetrics()
+	if err != nil {
+		m.logger.Warn("failed to update RAM metrics", zap.Error(err))
+		isError = true
+	}
+
+	// do not update metrics if some part is failed
```
antmat: it's quite strange decision - we update separate state, but we do not update the whole cached state

insonmnia/worker/metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+			case <-tk.C:
+				m.update(ctx)
+			}
+		}
+	}()
+}
+
+func (m *Handler) update(ctx context.Context) {
+	m.logger.Debug("updating hardware metrics")
+	isError := false
```
antmat: looks like multierror fits here

insonmnia/worker/metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+	dm := &sonm.DiskMetrics{
+		Total: info.AvailableBytes,
+		Free:  info.FreeBytes,
+	}
+
+	return dm, nil
+}
+
+func (m *Handler) updateRAMMetrics() (*sonm.RAMMetrics, error) {
+	meme, err := ram.NewRAMDevice()
```
antmat: meme? something like this

insonmnia/worker/gpu/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/sshaman1101/nvidia-docker/nvidia"
+	"go.uber.org/zap"
+)
+
+type nvidiaMetrics struct {
+	mu      sync.Mutex
+	devices []nvidia.Device
+}
+
+func newNvidiaMetricsHandler(ctx context.Context) (MetricsHandler, error) {
```
antmat: context only for logging? maybe pass logger instead?

insonmnia/worker/gpu/interface.go
```diff
@@ -38,6 +38,30 @@ func New(ctx context.Context, gpuType sonm.GPUVendorType, opts ...Option) (Tuner
...
+	Close() error
+}
+
+func NewMetricsHandler(ctx context.Context, gpuType sonm.GPUVendorType) (MetricsHandler, error) {
+	switch gpuType {
+	case sonm.GPUVendorType_RADEON:
+		return newRadeonMetricsHandler(ctx)
+	case sonm.GPUVendorType_NVIDIA:
+		return newNvidiaMetricsHandler(ctx)
+	default:
```
antmat: is it possible in real life? if no - maybe better return error?

insonmnia/worker/gpu/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+	if err := nvidia.Init(); err != nil {
+		ctxlog.G(ctx).Error("failed to load NVML", zap.Error(err))
```
3Hren: I'd rather return this as an error.

insonmnia/worker/gpu/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil, err
+	}
+
+	devices, err := nvidia.LookupDevices()
+	if err != nil {
+		ctxlog.G(ctx).Error("failed to collect devices via NVML", zap.Error(err))
+		return nil, err
```
3Hren: Here too.

insonmnia/worker/gpu/metrics.go
```diff
@@ -0,0 +1,114 @@
...
+
+type radeonMetrics struct {
+	mu      sync.Mutex
+	devices []DRICard
+}
+
+func newRadeonMetricsHandler(ctx context.Context) (MetricsHandler, error) {
+	devices, err := CollectDRICardDevices()
+	if err != nil {
+		ctxlog.G(ctx).Error("failed to collect DRI devices", zap.Error(err))
```
3Hren: Here too.

insonmnia/worker/metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+		m.mu.Lock()
+		m.lastState = &sonm.WorkerMetricsResponse{
+			GPUMetrics:  g,
+			CpuMetrics:  c,
+			DiskMetrics: d,
+			RamMetrics:  r,
+		}
+		m.mu.Unlock()
+
+		m.logger.Debug("new metrics received") // todo: remove
```
3Hren: Remove.

insonmnia/worker/metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+	return result, nil
+}
+
+func (m *Handler) updateCPUMetrics() (*sonm.CPUMetrics, error) {
+	loads, err := cpu.Percent(time.Second, false)
+	if err != nil {
+		return nil, err
+	}
+
+	u := float32(loads[0])
```
3Hren: Unchecked indexing.

insonmnia/worker/server.go
```diff
@@ -170,6 +172,8 @@ type Worker struct {
 
 	// Geolocation info.
 	country *geoip2.Country
+	// hardware metrics for various hardware types
```
```suggestion
	// Hardware metrics for various hardware types.
```

insonmnia/worker/server.go
```diff
@@ -365,6 +373,17 @@ func (m *Worker) setupPlugins() error {
 	return nil
 }
 
+func (m *Worker) setupMetrics() error {
+	h, err := metrics.NewHandler(m.ctx, m.cfg.Plugins.GPUs)
```
3Hren: Can we avoid passing context in New* functions, like here?

insonmnia/worker/server.go
```diff
@@ -2011,6 +2030,10 @@ func (m *Worker) AskPlanByTaskID(taskID string) (*sonm.AskPlan, error) {
 	return m.salesman.AskPlan(planID)
 }
 
+func (m *Worker) Metrics(ctx context.Context, req *sonm.WorkerMetricsRequest) (*sonm.WorkerMetricsResponse, error) {
+	return m.metrics.Get(), nil
```
3Hren: Seems like there is no way to break collecting if a client interrupts the request. Pass the context.

proto/worker.proto
```diff
@@ -202,3 +204,43 @@ message DebugStateReply {
 message PurgeTasksRequest {
     BigInt dealID = 1;
 }
+
+message WorkerMetricsRequest {}
+
+message GPUMetrics {
+    // pci bus id
```
3Hren: Please stick to go comments style, like "Id represents the PCI bus ID". Here and below.

proto/worker.proto
3Hren: This depends on how we will use them. If they are just for printing - then `map<string, NamedMetric>` is OK. Otherwise, it is better to have typed metrics like here.

insonmnia/worker/server.go
sshaman1101: Here `.Get()` is just returning cached value with lock, metrics collection already has the root ctx of the worker, so it will be canceled when a worker will be closed.

insonmnia/worker/server.go
sshaman1101: But we must terminate the metrics plugin on worker exit. Plugin stops collecting metrics when this context is canceling, like a lot of other components.

insonmnia/worker/metrics/metrics.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		merr = multierror.Append(merr, fmt.Errorf("failed to update disk metrics: %v", err))
+	}
+
+	ramMetrics, err := m.updateRAMMetrics()
+	if err != nil {
+		merr = multierror.Append(merr, fmt.Errorf("failed to update RAM metrics: %v", err))
+	}
+
+	if merr.ErrorOrNil() != nil {
```
3Hren:
```suggestion
	if err := merr.ErrorOrNil(); err != nil {
```

PR #1717, opened 2018-11-16
fix: allow to mount volumes only to privileged KYC level

antmat:


insonmnia/worker/server.go
```diff
@@ -1014,6 +1014,11 @@ func (m *Worker) taskAllowed(ctx context.Context, request *sonm.StartTaskRequest
 		return false, ref, err
 	}
 	if level < m.cfg.Whitelist.PrivilegedIdentityLevel {
+		volumes := request.GetSpec().GetContainer().GetVolumes()
```
```suggestion
		volumes := spec.GetContainer().GetVolumes()
```

insonmnia/worker/server.go
```diff
@@ -1014,6 +1014,11 @@ func (m *Worker) taskAllowed(ctx context.Context, request *sonm.StartTaskRequest
...
+		mounts := request.GetSpec().GetContainer().GetMounts()
```
```suggestion
		mounts := spec.GetContainer().GetMounts()
```

insonmnia/worker/server.go
```diff
@@ -1014,6 +1014,11 @@ func (m *Worker) taskAllowed(ctx context.Context, request *sonm.StartTaskRequest
...
+		if (volumes != nil && len(volumes) > 0) || (mounts != nil && len(mounts) > 0) {
```
```suggestion
		if len(volumes) > 0 || len(mounts) > 0 {
```
Since `len(nil)` gives 0.

PR #1724, opened 2018-11-19
feat: experimental QUIC protocol support

3Hren:
Tread lightly. What is it now? Speak, fool. At last. No one orders me around!

Introducing QUIC protocol support in most of the components. QUIC is
an experimental transport layer network protocol, which we use as an
alternative to TCP to achieve reliable communication between our
components.
The main reason to include this is to increase connection establishing
stability between any 2 peers, which are located under private networks.
QUIC is designed at top of UDP, which is more predictable and
investigated for NAT penetration, unlike we do for TCP in a hacky way.
Since QUIC allows to multiplex many connections into a single socket
there is no need to penetrate NAT for the same client-server pair
multiple times. After the first successful attempt, the punched hole will
be reused.

The Rendezvous server has been updated to support UDP hole punching using
QUIC protocol. It additionally exposes its gRPC server on the same UDP
port as for TCP, since it is possible to reuse ports in case of different
network protocols used. The wire-level left the same - gRPC.

Affected components: Worker, Node, Rendezvous. Both Node and Worker now
expose its gRPC services on both TCP and UDP sockets, the port number is
the same.

At last, Windows users now can utilize our NAT punching feature, since
multiplexed UDP communication does not require SO_REUSEPORT flag. So
relaying now isn't the only single option for Windows.

Note, that this feature is currently in an experimental stage. You can
try it by specifying "SONM_ENABLE_QUIC=true" environment while starting
Node server.

### DIFF
--- a/insonmnia/npp/puncher_quic.go
+++ b/insonmnia/npp/puncher_quic.go
@@ -1,0 +1,228 @@
+package npp
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"net"
+	"strings"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/lucas-clemente/quic-go"
+	"github.com/sonm-io/core/proto"
+	"github.com/sonm-io/core/util/xnet"
+	"go.uber.org/zap"
+	"gopkg.in/oleiade/lane.v1"
+)
+
+const (
+	transportProtocol = "quic"
+)
+
+type quicPuncher struct {
+	rendezvousClient *rendezvousClient
+	tlsConfig        *tls.Config
+
+	listenerChannel    chan connTuple
+	pendingConnections *lane.Queue
+	protocol           string
+
+	timeout time.Duration
+	log     *zap.SugaredLogger
+}
+
+func newQUICPuncher(rendezvousClient *rendezvousClient, tlsConfig *tls.Config, protocol string, log *zap.Logger) (*quicPuncher, error) {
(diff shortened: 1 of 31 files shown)


insonmnia/npp/puncher_quic.go
```diff
@@ -0,0 +1,233 @@
...
+
+	cfg := xnet.DefaultQUICConfig()
+
+	m.log.Warn("punchAddr")
+	session, err := quic.DialContext(ctx, udpConn, peerAddr, peerAddr.String(), m.tlsConfig, cfg)
+	if err != nil {
+		m.log.Warnf("punchAddr err: %s", err)
+		return nil, err
+	}
+	m.log.Warn("punchAddr done")
```
sshaman1101: warnings?

util/xnet/quic.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/lucas-clemente/quic-go"
+)
+
+func DefaultQUICConfig() *quic.Config {
+	return &quic.Config{
+		Versions: []quic.VersionNumber{
+			quic.VersionGQUIC39,
+			quic.VersionGQUIC43,
+			quic.VersionMilestone0_10_0,
```
sshaman1101: eto rabotaet

insonmnia/npp/puncher_quic.go
3Hren: Fixed.

PR #1736, opened 2018-11-21
fix: lock npm deps

antmat:
this PR locks all npm deps resulting in a more stable behaviour


3Hren:
lock at my package,
my package is amazing!

PR #1743, opened 2018-11-23
fix: add description for expose parameter and provide example 

antmat:


proto/container.proto
```diff
@@ -48,6 +48,10 @@ message Container {
     // ContainerRestartPolicy describes the restart policies of the container.
     ContainerRestartPolicy restartPolicy = 8;
     // Expose controls how container ports are exposed.
+    // Format is public_ip:public_port:private_port/protocol
+    // protocol can be "tcp", "udp", "sctp"
+    // if protocol is ommited "tcp" is implied
+    // if public_ip is ommited than port is being exposed on all available ips
```
3Hren:
```suggestion
    // If public_ip is omitted then the port is being exposed on all available IPs.
```

proto/container.proto
```diff
@@ -48,6 +48,10 @@ message Container {
     // ContainerRestartPolicy describes the restart policies of the container.
     ContainerRestartPolicy restartPolicy = 8;
     // Expose controls how container ports are exposed.
+    // Format is public_ip:public_port:private_port/protocol
```
3Hren:
```suggestion
    // Format is "public_ip:public_port:private_port/protocol".
```

proto/container.proto
```diff
@@ -48,6 +48,10 @@ message Container {
...
+    // protocol can be "tcp", "udp", "sctp"
```
3Hren:
```suggestion
    // Protocol can be "tcp", "udp", "sctp".
```

proto/container.proto
```diff
@@ -48,6 +48,10 @@ message Container {
...
+    // if protocol is ommited "tcp" is implied
```
3Hren:
```suggestion
    // If the protocol is omitted "tcp" is implied.
```

PR #1749, opened 2018-11-23
feat: show deal type in deal list cmd

### LINKED ISSUE #1445: CLI deal list - allocate BUY and SELL deals
Since we two times pull DWH - give my deals as a supplier, and give the deals as a buyer, you can select the type of deal - BUY (I'm the buyer) or SELL (I'm the supplier)

sshaman1101:
This commit adds the `type` field into the `sonmcli deal list` command.
The type indicates that the current user is selling or buying resources in that deal.

Closes #1445

### DIFF
--- a/cmd/cli/commands/printers.go
+++ b/cmd/cli/commands/printers.go
@@ -397,6 +397,16 @@ func dealsExpensesPerHour(addr common.Address, deals []*sonm.DWHDeal) (*sonm.Big
 	return sonm.NewBigInt(inAsks), sonm.NewBigInt(inBids)
 }
 
+func dealType(deal *sonm.Deal, addr common.Address) string {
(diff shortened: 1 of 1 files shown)


cmd/cli/commands/printers.go
```diff
@@ -397,6 +397,16 @@ func dealsExpensesPerHour(addr common.Address, deals []*sonm.DWHDeal) (*sonm.Big
 	return sonm.NewBigInt(inAsks), sonm.NewBigInt(inBids)
 }
 
+func dealType(deal *sonm.Deal, addr common.Address) string {
```
3Hren: What if BOTH conditions are true?

sshaman1101: fixed

PR #1756, opened 2018-11-27
feat: version checking for core components

3Hren:
Currently we often have a situation when our clients (both suppliers and customers) don't know whether a new version of the platform has been released. To avoid this we've done a simple console notification as a warning about whether it's time to update.

### DIFF
--- a/insonmnia/version/version.go
+++ b/insonmnia/version/version.go
@@ -1,0 +1,431 @@
+package version
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"strings"
+	"time"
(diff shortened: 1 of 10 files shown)


insonmnia/version/version.go
```diff
@@ -0,0 +1,437 @@
...
+	}
+
+	return &content, nil
+}
+
+func (m *VersionManager) isExpired(time time.Time) bool {
+	return time.Add(m.expireDuration).Before(m.clock())
+}
+
+// how to use it:
```
sshaman1101: Probably will be better to move usage info into the top of the file?

3Hren: Oh, it's just an ancient artifact. I'll drop it.

PR #1759, opened 2018-11-28
fix: do not hang on deal cancellation

antmat:
This PR introduces timeouts during deal cancellation in order not to hang forever if something(e.g. in docker) goes wrong


insonmnia/worker/salesman/options.go
```diff
@@ -4,6 +4,8 @@ import (
 	"crypto/ecdsa"
 	"errors"
 
+	"context"
```
3Hren: FMT!!!

PR #1788, opened 2018-12-14
Reduce delay on sonm-mon start

mou:
sonm-mon no longer waits to worker become available before rendering
Also produce warning message on losing connection to node or worker
Slightly tuned text layout


mou:
Merge please

PR #1810, opened 2018-12-20
fix: migration with proper multisig

antmat:
subj, now we use the proper multisig contract

### DIFF
--- a/blockchain/source/migrations/2_deploy_v1.js
+++ b/blockchain/source/migrations/2_deploy_v1.js
@@ -149,8 +149,6 @@ async function deploySidechain (deployer, network, accounts) {
     await ahm.write('profileRegistryAddress', pr.address, { gasPrice: 0 });
     await ahm.write('oracleUsdAddress', oracle.address, { gasPrice: 0 });
     await ahm.write('gatekeeperSidechainAddress', gk.address, { gasPrice: 0 });
-    await ahm.write('multiSigAddress', multiSig.address, { gasPrice: 0 });
-    // compatibility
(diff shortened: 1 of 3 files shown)


blockchain/source/migrations/2_deploy_v1.js
```diff
@@ -134,21 +135,25 @@ async function deploySidechain (deployer, network, accounts) {
...
+
     await ahm.write('blacklistAddress', bl.address, { gasPrice: 0 });
     await ahm.write('marketAddress', market.address, { gasPrice: 0 });
     await ahm.write('profileRegistryAddress', pr.address, { gasPrice: 0 });
     await ahm.write('oracleUsdAddress', oracle.address, { gasPrice: 0 });
     await ahm.write('gatekeeperSidechainAddress', gk.address, { gasPrice: 0 });
     await ahm.write('multiSigAddress', multiSig.address, { gasPrice: 0 });
     // compatibility
-    await ahm.write('migrationMultSigAddress', multiSig.address, { gasPrice: 0 });
+    await ahm.write('migrationMultiSigAddress', multiSig.address, { gasPrice: 0 });
```
sokel: you're write the same address for `migrationMultiSigAddress` and `multiSigAddress` ?

antmat: yep, that's right

antmat: ok, maybe it is not a good idea, if we still will have several multisigs in future

sokel: ok, it's really right.

PR #1814, opened 2018-12-24
chore: performed live migration of devices storage

antmat:


antmat:
such logs, much clear, wow

PR #1826, opened 2019-01-14
fix: enable interceptor for REST interface on DWH

sshaman1101:
At least we also should notify a user about the "out of sync" state.

### DIFF
--- a/insonmnia/dwh/server.go
+++ b/insonmnia/dwh/server.go
@@ -152,7 +152,11 @@ func (m *DWH) serveHTTP() error {
 		m.mu.Lock()
 		defer m.mu.Unlock()
 
-		options := []rest.Option{rest.WithLog(m.logger)}
+		options := []rest.Option{
+			rest.WithLog(m.logger),
+			rest.WithInterceptors(m.unaryInterceptor),
+		}
+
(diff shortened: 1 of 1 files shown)


3Hren:
Seems no reason for chaining.

blockchain/api.go
```diff
@@ -1498,6 +1498,27 @@ func (api *BasicEventsAPI) fetchAndProcessLogs(ctx context.Context, filter simpl
...
+	// just push NoEvent data.
+	if len(logs) == 0 {
+		for blockNumber := filter.FromBlock; blockNumber <= filter.ToBlock; blockNumber++ {
+			block, err := api.client.BlockByNumber(ctx, big.NewInt(0).SetUint64(blockNumber))
+			if err != nil {
+				ctxlog.S(ctx).Warn("failed to get block timestamp", zap.Uint64("block", blockNumber), zap.Error(err))
+				continue
+			}
+
+			receiver <- &Event{
```
antmat: this could break code which rely on the fact that filter returns event of only one kind. Please check the usage everywhere.

antmat: Also you can add some parameter to filter to support different types of behaviour

PR #1829, opened 2019-01-16
feat: allow to share admin capabilities for workers

3Hren:
Sometimes it is useful to provide additional capabilities to your Worker for other people. For example, temporarily allow to SSH into your host system to perform debug.
This addition allows performing such operations.

This is an experimental feature, so no CLI changes done. One can utilize autocli to be able to use this.


3Hren:
Need to think more.

PR #1835, opened 2019-01-21
feat: worker inspection

3Hren:
Currently, if something goes wrong on a Worker we're still able to gain
access to the remote computer via SSH to find of what's going on there.
In the future, due to obvious trusted platform limitations, we will
lose those capabilities, since there will be no direct access to the machine.
This leads us to a situation where we still need to collect debugging
information, but more likely in a post-mortem way.
This feature introduces the inspection service on the Worker side, which
is capable to collect various runtime and history information and
provide it to the subject that is explicitly allowed by Worker's admin.
New methods include config, list of open files, network information,
various Docker info and of course - logs. Now it is possible both for
Worker's admin and allowed users to subscribe to Worker logs.
Also, Worker's API now allows to add and to remove additional
capabilities like inspection service access or SSH'ing into the host.

[asciicast](https://asciinema.org/a/Mj2NQxt23pdkmfexyepJHjY3K)


insonmnia/auth/auth.go
```diff
@@ -155,18 +160,138 @@ type transportCredentialsAuthorization struct {
...
+		entry.NotifySubscribers()
+	}
+	delete(m.entries, addr)
+}
+
+func (m *AnyOfTransportCredentialsAuthorization) Authorize(ctx context.Context, request interface{}) error {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	peerInfo, err := FromContext(ctx)
```
antmat: move upper

insonmnia/auth/auth.go
```diff
@@ -155,18 +160,138 @@ type transportCredentialsAuthorization struct {
...
+	m.mu.Lock()
+	defer m.mu.Unlock()
 
-	switch authInfo := peerInfo.AuthInfo.(type) {
-	case EthAuthInfo:
-		if equalAddresses(authInfo.Wallet, a.ethAddr) {
-			return nil
+	for _, entry := range m.entries {
+		if entry.IsExpired() {
+			entry.NotifySubscribers()
```
antmat: either comment why you don't delete entry or delete  it if possible

3Hren:
Fixed.

cmd/cli/commands/worker.go
```diff
@@ -244,3 +250,102 @@ var workerDebugStateCmd = &cobra.Command{
...
+				return err
+			}
+
+			cmd.Printf("%s", message.GetMessage())
+		}
+	},
+}
+
+var workerAddCapabilityCmd = &cobra.Command{
+	Use:   "add-capability ETH_ADDR CAPABILITY TTL",
```
sshaman1101: In other commands, we're describing required params as `<param>`.
I'd suggest allowing to set TTL as `time.Duration` which I think more user-friendly. BTW, can we use text IDs for the Capability instead of numbers?

insonmnia/worker/server.go
```diff
@@ -2044,6 +2093,32 @@ func (m *Worker) Metrics(ctx context.Context, req *sonm.WorkerMetricsRequest) (*
...
+		return nil, fmt.Errorf("unknown scope: %d", request.GetScope())
+	}
+
+	return &sonm.WorkerAddCapabilityResponse{}, nil
+}
+
+func (m *Worker) RemoveCapability(ctx context.Context, request *sonm.WorkerRemoveCapabilityRequest) (*sonm.WorkerRemoveCapabilityResponse, error) {
+	switch request.GetScope() {
+	case sonm.CapabilityScope_CAPABILITY_NONE:
+	case sonm.CapabilityScope_CAPABILITY_INSPECTION:
```
sshaman1101: but we also have SSH...

3Hren: Where?

cmd/cli/commands/worker.go
3Hren: raw int produces nanosecods. Are you sure this is what we want?

PR #1857, opened 2019-02-19
feat: benchmark support for GPU type

antmat:
this PR adds benchmarks for GPU type - nvidia or amd only, which is usefull when specific card type is required.

### DIFF
--- a/insonmnia/worker/server.go
+++ b/insonmnia/worker/server.go
@@ -1778,9 +1778,9 @@ func (m *Worker) getBenchValue(bench *sonm.Benchmark, device interface{}) (uint6
 		return gpuDevice.GetMemory(), nil
 	}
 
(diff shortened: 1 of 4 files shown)


insonmnia/worker/server.go
```diff
@@ -1778,6 +1778,26 @@ func (m *Worker) getBenchValue(bench *sonm.Benchmark, device interface{}) (uint6
 		return gpuDevice.GetMemory(), nil
 	}
 
+	if bench.GetID() == benchmarks.GPUAMD {
+		if !isGpu {
+			return uint64(0), fmt.Errorf("invalid device for GPUMem benchmark")
```
sshaman1101: so copy-paste!

sshaman1101:
I made it work.
20190219-2102-5e4yq

PR #1858, opened 2019-02-26
Sensible defaults for brute-force optimisation

mou:
Default value are too high


3Hren:
For some unknown reasons checks were failed.

PR #1869, opened 2019-03-25
fix: gomod reuseport replacement

sokel:
This pr fixes incorrect `github.com/libp2p/go-reuseport` dependency definition.
It's replaces automatically after every build.


sokel:
In other case, i have tests with newer version of go. 
This have any problems with checksums, I do that we need to update project go version to latest stable and update dependencies.

PR #1884, opened 2019-06-25
fix: temp fix for `alien` deals bound to ask-plans

antmat:
This one need to be tested. Also there is still also a probability to get `alien` order which need to be fixed in blockchain API in general, not the salesman

### DIFF
--- a/insonmnia/worker/salesman/salesman.go
+++ b/insonmnia/worker/salesman/salesman.go
@@ -532,6 +532,20 @@ func (m *Salesman) checkDeal(ctx context.Context, plan *sonm.AskPlan) error {
 		if err := m.RemoveAskPlan(ctx, plan.GetID()); err != nil {
 			return fmt.Errorf("failed to remove ask plan %s with closed deal: %s", plan.GetID(), err)
 		}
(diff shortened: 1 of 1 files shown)


insonmnia/worker/salesman/salesman.go
```diff
@@ -532,6 +532,20 @@ func (m *Salesman) checkDeal(ctx context.Context, plan *sonm.AskPlan) error {
...
+		defer m.mu.Unlock()
+		// We have a deep copy of ask-plan, so refetch original one
+		plan, ok := m.askPlans[plan.ID]
+		if !ok {
+			return fmt.Errorf("failed to fetch ask plan by id %s", plan.ID)
+		}
+		plan.DealID = nil
+		if err := m.askPlanStorage.Save(m.askPlans); err != nil {
+			return err
+		}
```
3Hren: Better to `return nil` here explicitly, because there is a lock taken here with defer. If someday we add code after this if/else with taking a mutex we can occasionally be deadlocked.

3Hren: Even better to unlock explicitly, but who cares.

PR #1886, opened 2019-07-01
feat: ability to extract bid and ask tags from DWH

sshaman1101:
¯\(ツ)/¯


3Hren:
What about travis?

sshaman1101:
@3Hren fixed.

PR #1895, opened 2019-07-25
feat: remote secure shell

3Hren:
This commit allows exposing a server on a host, which act as a remote secure shell, executing allowed commands within the security policy.

- [x] Restricted PTY.
- [x] Load seccomp profiles from the filesystem (for multiple commands).
- [x] Restrict capabilities.
- [x] Set `PR_SET_NO_NEW_PRIVS`.
- [x] Wrap server side with gRPC or SSH.
- [x] PTY on the client-side.
- [x] Penetrate NAT.
- [x] Reconnect when failed.
- [x] Support pipes.
- [x] ~Stream stdin~.
- [x] Prevent virtual priv escape.
- [x] Prevent change limits (rlimits).
- [x] Select policy in secexec, not inside the server.
- [x] List of allowed commands.
- [x] Drop capabilities selectively.
- [x] Executable checksum verify.
- [x] ACL.
- [x] Integrate into QOS/spy.


secexec/src/exec.rs
```diff
@@ -0,0 +1,110 @@
...
+        } else {
+            for rule in rules {
+                ctx.add_rule(syscall, Action::Allow, Some(*rule))?;
+            }
+        }
+    }
+
+    ctx.load()?;
+
+    //    println!("[!] PREPARE YOUR ANUS");
```
sshaman1101: 🤔🤔🤔

PR #1896, opened 2019-07-29
feat: add bid and ask tags to dwh deal

oopcode:
This PR adds `AskTag` and `BidTag` fields to DWH Deal representation.


oopcode:
You f****rs better review this ***t.

PR #1901, opened 2019-08-21
feat: allow to specify pwd in secexec

3Hren:
Also the usage of "secexec" has been slightly changed: now the executable and its arguments must be specified after "--".
For example: secexec /etc/secexec -- /bin/ls -lith


spyinfo:
good

PR #1907, opened 2020-04-13
fix: more nvidia related .so files for volumes

sshaman1101:
Found out that we're skipping a lot of useful nvidia's .so files.
This PR adds all of them to the docker volume.
Based on the content of NVIDIA-Linux-x86_64-440.82.

To keep the list up-to-date in the future:
1. download the latest version on drivers;
2. unpack them without installing: `./NVIDIA-Linux-x86_64-XXX.YY.run --extract-only`;
3. list .so and update the list in `insonmnia/worker/gpu/nvidia_tuner.go`; Don't forget to remove versions and .0 from file names;
4. ???
5. Congratulation, you're awesome open-source contributor;


3Hren:
I express my LGBT

